PR #1, opened 2014-06-24
Add "status" command to get Scrapinghub services statuses

redapple:


pablohoffman:
Needs rebase.

+0 for merging. I don't see myself using this command, but if others find it useful I don't object.

pablohoffman:
I'm more inclined to merge a `status` command now, but this one needs to be redone to use statuspage.io instead of pingdom. Perhaps statuspage.io have a more convenient public API?

PR #10, opened 2014-12-27
Fix netrc permissions

bertinatto:
Here is a PR to make sure that the netrc file (created by login command) can only be accessed by its owner (0600 permissions).


bertinatto:
I hadn't tested it on Windows, sorry. I just checked in some fixes.

Just a note: Because Windows uses a different permission model, "0600" mode has no effect in it. `heroku-cli` also has this behavior on Windows (created `_netrc` file has default permissions).

PR #14, opened 2015-04-22
Add "deploy_egg" command

josericardo:
Ported from the useful_scripts repo:

https://github.com/scrapinghub/useful_scripts/blob/master/scripts/build_and_deploy_egg_from_source.sh


eliasdorneles:
Thanks, @josericardo, looks good to me! +1

pablohoffman:
:+1: @josericardo 

probably worth removing the one from useful_scripts now.

eliasdorneles:
Hang on, let's wait a bit before deleting it from useful_scripts: there are other scripts depending on that one. It's probably better to finish porting the remaining scripts, before deleting it (or change it to use `shub`).

PR #15, opened 2015-04-22
Use hifens instead of underscore for command names

josericardo:
Easier to type?


pablohoffman:
which command uses this separator?

PR #16, opened 2015-04-27
Add "deploy-reqs" command

josericardo:
A port from the deploy_requirements script from useful_scripts.

It's a wrapper around the deploy-egg command. 
It downloads, builds and deploys dependencies read from a requirements.txt file.

I couldn't think of a simple way to test it, since it has to download the packages. Besides from that, it's just a wrapper around deploy-egg, which has tests.


pablohoffman:
looks good, how about `deploy-reqs` for a shorter command name?

josericardo:
deploy-reqs is better, updated :)

pablohoffman:
Looks good. For testing, what about creating (in an ephemeral temporary dir) a fake requirements.txt and 2 fake libraries in the same dir, and make sure those two are deployed?. Or at least they invoke deploy_egg the way it's supposed to.

josericardo:
Test added :)

pablohoffman:
@josericardo can you rebase?. It doesn't merge in the current state.

josericardo:
Sure @pablohoffman, done :)

PR #17, opened 2015-05-05
Add "download-eggs" command

josericardo:
Downloads all the project's eggs as a zip file. 

Very simple, just hits the project's eggs bundle url and saves it.


pablohoffman:
How about `fetch-eggs` for the command name?. Just because it's shorter.

josericardo:
Ok, done.

PR #24, opened 2015-06-13
Add version command

eliasdorneles:
Hey fellows,

Here is a proposal adding the `__version__` variable in `__init__.py` (updated automatically when doing the release with `bumpversion`) and it also adds a `shub version` command which will give the current version.

Does this look good?


eliasdorneles:
I think we're ready for a release after this one.

I intended to do the release now, but realized I lack the necessary pypi access. /cc @pablohoffman @dangra

dangra:
As far as I see travis-ci is configured to do the release on new tags.

dangra:
I mean, just push the changes and the tags added by bumpversion `git push origin master --tags`

eliasdorneles:
@dangra ahh, that's cool, didn't know that, thank you!

well, then I'll merge this PR and do the release (there are a bunch of things depending on it already) -- if there is a problem we do a pach release.

eliasdorneles:
Ops, I just realized that the tests are broken, the Travis CI build is not running the tests and neither does tox locally -- they're both reporting all green because no tests are being collected. :)

I shall fix that before doing the release.

pablohoffman:
Ooops!. Good catch @eliasdorneles

PR #25, opened 2015-06-14
Fix tox config to run tests

eliasdorneles:
So, the build will fail for this one, there are some broken tests that weren't running before.

I think we can merge this anyway, since it will reflect the real state of the build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist
 *.egg-info
 *.pyc
 .eggs/
+.tox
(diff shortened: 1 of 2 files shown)


eliasdorneles:
I'm looking how to fix the broken tests, it seems to be related to a sample git repo in the test fixtures that wasn't commited.
Not sure if it's even possible to commit a git repo inside a git repo. I think we can put it into a zip and extract it before each test run.

Thoughts?

pablohoffman:
Sounds good @eliasdorneles

PR #27, opened 2015-06-14
Fix tests for python3 using decode() on pipe output

eliasdorneles:
Hey folks,

So, this needs more testing, it just adds some tricks to fix the current tests for Python 3, I'm not sure how much this means for Python 3 compatibility yet.

This basically just adds `.decode()` to the places where are expected bytes in Python 3 (pipe outputs) -- plus some minor refactorings.

The encoding should be `ascii` for all outputs, not sure if I should be explicit about it.

@kmike any advice here? :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: python
 python: 2.7
 env:
 - TOXENV=py27
+- TOXENV=pypy
+- TOXENV=py33
+- TOXENV=py34
 install:
 - "./.travis-workarounds.sh"
 - pip install -U tox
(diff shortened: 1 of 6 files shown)


kmike:
`.decode()` and `.encode()` use `sys.getdefaultencoding()` encoding, which is usually ASCII. For pipes it may be better to use sys.stdout.encoding / sys.stdin.encoding with some fallback for the cases these encodings are not set (utf8?).

pablohoffman:
who should merge this?

eliasdorneles:
@pablohoffman Ahm, allow me to apply Mike's advice first, and then you or @josericardo can review & merge.

You'd have to run the tests locally with `tox`, which will run them for python 3.
Travis is only running them for Python 2, perhaps I should enable python 3 tests as part of this PR?

eliasdorneles:
updated and rebased!

PR #30, opened 2015-06-17
Add rudimentary end to end tests

josericardo:
Our subcommand tests bypass the click wrappers. I think it's a good idea
to have at least these very simple end to end tests (although not very useful) just to have a little more confidence that no silly bugs have gone unnoticed.

For example, this commit https://github.com/scrapinghub/shub/commit/e73c4e34bceaf8a3469a42a0167306a46e8ef7d7 didn't allow the subcommands to be run, although ```shub```, ```shub --version``` and ```shub version``` were running fine (I tried those when testing the above PR =/).

### DIFF
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -4,8 +4,11 @@
 import unittest
 from click.testing import CliRunner
 from shub import tool
+import os
 
 
+@unittest.skipUnless(os.getenv('USING_TOX'),
+                     'End to end tests only run via TOX')
 class ShubEndToEndTests(unittest.TestCase):
     def setUp(self):
         self.runner = CliRunner()
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+# coding=utf-8
+
+import unittest
+from click.testing import CliRunner
+from shub import tool
+
+
+class ShubEndToEndTests(unittest.TestCase):
+    def setUp(self):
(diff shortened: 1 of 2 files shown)


eliasdorneles:
+1

Only one question: the tests pass with `py.test`, however running the tests with nosetests I get an error:

```FAIL: test_deploy_isnt_broken (tests.test_end_to_end.ShubEndToEndTests)
----------------------------------------------------------------------
  File "/home/elias/hacking/scrapinghub/shub/tests/test_end_to_end.py", line 34, in test_deploy_isnt_broken
    self.assertTrue('requires scrapy' in output, error)
AssertionError: Unexpected output: Error: no Scrapy project found in this location
```

Can you reproduce that or is it my crazy environment?

josericardo:
The ```deploy``` test is a little fragile. It's defined to work on tox (where scrapy isn't installed). If you run your tests in an environment that has scrapy installed it will break. Should I make it more flexible? I think we all should be running the tests via tox (I wasn't until I stumbled on this error).

eliasdorneles:
Hm, the test suite should run outside of tox -- tox is slow to give feedback, it's not meant to be used while refactoring for example.

If this is a test to run only via tox, let's run it only in tox, then. =)
We can add `@unittest.skipIf(os.getenv('USING_TOX'))` to the tests (I think the whole end-to-end suite) and update `tox.ini` to set the env variable like this:

```
[testenv]
setenv =
    USING_TOX=1
```

What do you think?

josericardo:
I like it, pushed :)

PR #36, opened 2015-06-19
Deploy falling back to standard login credentials

### LINKED ISSUE #35: 'shub login' doesn't work for 'shub deploy'
```shub deploy``` only looks for credentials in the project's scrapy.cfg and ~/.scrapy.cfg files.

Does it make sense to make it work with .netrc (shub login) and $SHUB_APIKEY credentials?

### PR DESCRIPTION
**Pull request by josericardo**
It was looking only for scrapy.cfg files. Now it also looks for the
~/.netrc file and $SHUB_APIKEY.

Fixes #35

### DIFF
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -16,8 +16,11 @@ from scrapy.utils.project import inside_project
 from scrapy.utils.python import retry_on_eintr
 from scrapy.utils.conf import get_config, closest_scrapy_cfg
 
+from shub import login
+
 from shub.click_utils import log, fail
-from shub.utils import make_deploy_request, pwd_hg_version, pwd_git_version
+from shub.utils import (make_deploy_request, pwd_hg_version, pwd_git_version,
+                        find_api_key)
 
 
 _SETUP_PY_TEMPLATE = """\
@@ -149,8 +152,13 @@ def _get_auth(target):
     # try netrc
     try:
         host = urlparse(target['url']).hostname
-        a = netrc.netrc().authenticators(host)
-        return (a[0], a[2])
+        auth = netrc.netrc().authenticators(host)
+        if auth:
+            return (auth[0], auth[2])
+
+        if host.endswith(login.SH_TOP_LEVEL_DOMAIN):
+            return (find_api_key(), '')
+
     except (netrc.NetrcParseError, IOError, TypeError):
         pass
 
--- a/shub/login.py
+++ b/shub/login.py
@@ -1,5 +1,10 @@
-import os, re, click
+import os
+import re
+import click
 from shub.utils import get_key_netrc, NETRC_FILE
+
+SH_TOP_LEVEL_DOMAIN = 'scrapinghub.com'
+
 
 @click.command(help='add Scrapinghug API key into the netrc file')
 @click.pass_context
@@ -12,10 +17,12 @@ def cli(context):
             NETRC_FILE,
             os.O_CREAT | os.O_RDWR | os.O_APPEND, 0600)
         with os.fdopen(descriptor, 'a+') as out:
-            line = 'machine scrapinghub.com login {0} password ""\n'.format(key)
+            line = 'machine {0} login {1} password ""\n'.format(
+                SH_TOP_LEVEL_DOMAIN, key)
             out.write(line)
     else:
         context.fail('Invalid key')
+
 
 def is_valid_key(key):
     return bool(re.match(r'[A-Fa-f\d]{32}$', key))

### CODE REVIEW

**Comment by pablohoffman** [comment]
Rather than adding more code to `shub.deploy` to read auth from other sources, I prefer to steer users into using the login/logout workflow.

So, when users try a command ( `deploy`, `fetch-eggs`, etc) the tool should say it's not logged in and ask the user to log in.

Then, `shub login` should provide some convenient shortcut to read credentials (ie. API key) from  `scrapy.cfg` (and other places?), but this code to read auth from multiple places is only in `shub.login`.

Example to illustrate:

```
$ shub deploy
Error: Not logged in. Please login first with: shub login
```

Then...

Option A: same prompt we show now, but reading default value (for API key) from adjacent `scrapy.cfg`:

```
$ shub login
Insert your Scrapinghub API key [c72c12a4ac15bbedf6ee0f1d5161263c]: 
```

Option B: be more explicit about `scrapy.cfg`:

```
$ shub login
We found this API key in your scrapy.cfg: c72c12a4ac15bbedf6ee0f1d5161263c. Use this API key to log in?
```

**Comment by eliasdorneles** [comment]
@pablohoffman +1, I liked Option A!

@josericardo can you implement that?

Let's make a patch release after that, this issue is already hitting some customers.

**Comment by josericardo** [comment]
@eliasdorneles sure

+1 for option A, but I'm wondering how unusual it is to have different credentials for different projects. I mean, it may get a little annoying to login and logout frequently if the user is switching between projects with different credentials. If that's the case, should we compare the login credentials to the project's credentials and ask the user to choose one of them? Or is it rare enough to be ignored?

**Comment by josericardo** [comment]
Actually, this is a more general concern, right? Should the user be asked the project id, for example, if he is on a scrapy project folder? Food for thought for future releases, maybe.

**Comment by pablohoffman** [comment]
The project id should remain in scrapy.cfg, it's only the auth that is being moved.

You pretty much never need to change API. The API key identifies your user at Scrapinghub, and that user may/will have access to multiple projects. This is similar to how git+github works, auth is stored in a single file (usually `~/.ssh/id_rsa`).

**Comment by josericardo** [comment]
Ok. I mentioned the project id more from a convenience standpoint for the other commands (not deploy), like just typing ```shub deploy-reqs requirements.txt``` could be enough if I'm on the project folder.

**Comment by eliasdorneles** [comment]
+1 to update `deploy-reqs` and `deploy-egg` to try to get the project id from current project and be more consistent with `deploy` behavior.

**Comment by eliasdorneles** [comment]
@josericardo ^^ that'd be a separate PR/issue. :)

**Comment by josericardo** [comment]
@eliasdorneles yeah, sorry for mixing things. As I said, food for thought for future releases.

**Comment by chekunkov** [comment]
This change doesn't make existing customers happy. At least it confuses them a lot because they already have their keys in `~/.scrapy.cfg` and suddenly after switching to shub 1.3.1 deploy breaks - they are writing private messages about that and we are explaining that they should run `shub login`. Was it actually a good idea to restrict auth to `~/.netrc` file without keeping backwards compatibility with `~/.scrapy.cfg` which was used for years? Moving key from `~/.scrapy.cfg` to `~/.netrc` doesn't provide benefits to average customers, it's just confuses them and wastes their time.

**Comment by pablohoffman** [comment]
Agree with Alex. We should have provided enough guidance to move to the new
way without having to check a forum or ask. Like with a simple prompt where
users can hit Yes. That was the idea with the default value taken from
scrapy.cfg (which should also cover the home file). But extra care should
have been taken to users not aware that they need to login now.

**Comment by chekunkov** [comment]
Also @josericardo was right about inconvenience in case of multiple credentials. Let me provide an example. We have custom service built for customer and we used `scrapy deploy` for deploy, so our scrapy.cfg in the project looked like

```
[settings]
default = project.settings

[deploy]
url = https://dash.scrapinghub.com/api/scrapyd/
project = <PROJECT_ID>
version = GIT

...

[deploy:scrapyrt-service]
project = scrapyrt-service
url = http://service.domain
username = <SOME_USERNAME>
password = <SOME_PASSWORD>
```

So it was possible to deploy by using just `scrapy deploy scrapyrt-service`. After switch to `shub` (version 1.2.0) it still was possible to `shub deploy scrapyrt-service`. After upgrade to `1.3.1` it just doesn't work, `shub` is hardcoded to use netrc config and looks only for `scrapynghub.com` entry there, so now there's no way to reuse deploy functionality from `shub` for other purposes.

Any ideas how to overcome this @pablohoffman @josericardo? I don't really want to backport `scrapy deploy` from outdated scrapy version to the project only for this service, but it seems like now `shub` isn't well suited for anything other than dash.scrapinghub.com deploy.

**Comment by josericardo** [comment]
I'm sorry that this happened @chekunkov. I don't think that preserving the old deploy behavior you mentioned (allowing multiple targets/credentials) is a bad idea. I can make a PR to get this support back.

**Comment by chekunkov** [comment]
@josericardo you don't have to be sorry. Probably we need to wait for @pablohoffman decision, I can accept answer "shub is intended to be used only with scrapinghub.com", because using shub in such way is a bit of a hack and that can be proved by checking output of shub when it's used with those credentials:

```
$ shub deploy scrapyrt-service
Packing version 70f567c-master
Deploying to Scrapy Cloud project "scrapyrt-service"
{"status": "ok", "version": "70f567c-master"}
Run your spiders at: https://dash.scrapinghub.com/p/scrapyrt-service/
```

There is no Scrapy Cloud project "scrapyrt-service" and link is meaningless in this case.

chekunkov:
hey folks, the way we use `.netrc` confuses me a bit, I shared my thoughts [here](https://github.com/scrapinghub/shub/pull/40/files#r48473811), please share yours.

Issue #34, opened 2015-06-19
Add authentication instructions when the authentication fails


josericardo:
Today we only display "Authentication failed". 

The docs also are outdated:

http://doc.scrapinghub.com/shub.html

josericardo:
Removing from scrapinghub docs here: https://github.com/scrapinghub/doc.scrapinghub.com/pull/72

josericardo:
Fixed by #37

PR #37, opened 2015-06-19
Raising exception on utils#find_api_key

### LINKED ISSUE #32: The command `fetch-eggs` doesn't fail on auth error
The example below shows how `shub fetch-eggs` creates a zip file without the actual zip content due to missing auth key.

```
$ shub fetch-eggs 1261X
Downloading eggs to eggs-1261X.zip
$ file eggs-1261X.zip 
eggs-1261X.zip: ASCII text, with no line terminators
$ export SHUB_APIKEY=xxxxxxxb31
$ shub fetch-eggs 1261X
Downloading eggs to eggs-1261X.zip
$ file eggs-1261X.zip
eggs-12616.zip: Zip archive data, at least v2.0 to extract
```

josericardo:
NoCredentialsException is a ClickException, which is handled nicely by
click, displaying the error msg and returning an exit code != 0.

Fixes #32 and #34

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -36,10 +36,20 @@ def missing_modules(*modules):
 
 
 def find_api_key():
-    """Finds and returns the Scrapy Cloud APIKEY"""
+    """Finds and returns the Scrapy Cloud APIKEY
+
+    Raises:
+        ClickException: if no API key is found
+    """
     key = os.getenv("SHUB_APIKEY")
     if not key:
         key = get_key_netrc()
+
+    if not key:
+        err = ("Your credentials haven't been defined.\n"
+               "Use 'shub login' or set the SHUB_APIKEY environment variable.")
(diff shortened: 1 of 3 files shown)


pablohoffman:
Why do we need the custom class?. Why not just raise ClickException with the error message?

josericardo:
As a rule, I try not to raise generic exceptions. I think the code gets a little more expressive. Should I remove it?

pablohoffman:
I don't think there's a need for a custom exception in this case, and I prefer to keep the code simpler without them.

pablohoffman:
The first need for a custom exception class would come when we need to raise the *same* exception from multiple places, and in that case the exception message ("Use 'shub login' or ...") should probably be in the class definition, not passed at exception creation.

josericardo:
You're right. Updated :)

PR #38, opened 2015-06-19
Raise ClickException in case of request errors

### LINKED ISSUE #32: The command `fetch-eggs` doesn't fail on auth error

josericardo:
Better than handling boolean returns. As it has already proved to be prone
to errors (not handling the return values).

Fixes #32

### DIFF
--- a/shub/fetch_eggs.py
+++ b/shub/fetch_eggs.py
@@ -1,8 +1,10 @@
 import click
 import requests
+from click import ClickException
 
 from shub.utils import find_api_key
 from shub.click_utils import log
+from shub.exceptions import AuthException
 
 
 @click.command(help="Download a project's eggs from the Scrapy Cloud")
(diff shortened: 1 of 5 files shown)


shub/fetch_eggs.py
```diff
@@ -12,6 +14,8 @@ def cli(project_id):
     url = "https://dash.scrapinghub.com/api/eggs/bundle.zip?project=%s" % project_id
     rsp = requests.get(url=url, auth=auth, stream=True, timeout=300)
 
+    assert_response_is_valid(rsp)
```
josericardo: Fixes #32

shub/utils.py
```diff
@@ -48,7 +51,6 @@ def get_key_netrc():
     if key:
         return key
 
-
 def make_deploy_request(url, data, files, auth):
```
josericardo: The other commands besides ```fetch-eggs``` use this method to make requests. That's why I handled the status code also here.

PR #40, opened 2015-06-26
'shub deploy' using the apikey set by 'shub login'

### LINKED ISSUE #35: 'shub login' doesn't work for 'shub deploy'

josericardo:
As discussed on #36, all commands should be in the login/logout workflow.

Fixes #35 

I'll make another PR to facilitate the login reading api keys in scrapy.cfg files.

### DIFF
--- a/shub/auth.py
+++ b/shub/auth.py
@@ -1,0 +1,48 @@
+import os
+import netrc
+
+from os.path import expanduser
+
+from click import ClickException
+
+OS_WIN = True if os.name == 'nt' else False
+NETRC_FILE = expanduser('~/_netrc') if OS_WIN else expanduser('~/.netrc')
+
+
+def find_api_key():
+    """
+    Raises:
+        ClickException: if no credentials are found
+    """
+    key = os.getenv("SHUB_APIKEY")
+    if not key:
+        key = get_key_netrc()
+
+    if not key:
+        err = 'Not logged in. Please login first with: shub login'
+        raise ClickException(err)
+
+    return key
+
+
+def get_key_netrc():
+    """Gets the key from the netrc file"""
+    try:
+        info = netrc.netrc(NETRC_FILE)
+    except IOError:
+        return
+    try:
+        key, account, password = info.authenticators("scrapinghub.com")
+    except TypeError:
+        return
+    if key:
+        return key
+
+
+def write_key_netrc(key):
+    descriptor = os.open(
+        NETRC_FILE,
+        os.O_CREAT | os.O_RDWR | os.O_APPEND, 0o600)
+    with os.fdopen(descriptor, 'a+') as out:
+        line = 'machine scrapinghub.com login {0} password ""\n'.format(key)
+        out.write(line)
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -2,10 +2,9 @@ import os
 import sys
 import glob
 import time
-import netrc
 import shutil
 import tempfile
-from urlparse import urlparse, urljoin
+from urlparse import urljoin
 from subprocess import check_call
 
 import click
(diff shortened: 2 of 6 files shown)


shub/auth.py
```diff
@@ -0,0 +1,48 @@
...
+OS_WIN = True if os.name == 'nt' else False
+NETRC_FILE = expanduser('~/_netrc') if OS_WIN else expanduser('~/.netrc')
+
+
+def find_api_key():
+    """
+    Raises:
+        ClickException: if no credentials are found
+    """
+    key = os.getenv("SHUB_APIKEY")
```
josericardo: Although we want to keep users in the login/logout workflow, I think $SHUB_APIKEY is convenient for automated scripts, right? It can be used to deploy projects from CI, for example.

josericardo:
Should we display a deprecation warning if the there's a ```user``` key on the scrapy.cfg file? Is it used anywhere else?

shub/auth.py
pablohoffman: Yeah, it's fine to leave support for `SHUB_APIKEY` as long as the "logged in" API superseedes it.

The `SHUB_APIKEY` env var should also be used to populate the default API key for login.

pablohoffman:
@josericardo it's also used by scrapyd deploy tool (which is where scrapy.cfg deploy configuration comes from). I would however issue a warning because leaving the API keys / credentials on the project's `scrapy.cfg` is almost often a bad idea. 

However, if the API key is stored in `~/.scrapy.cfg` it's not bad.

shub/auth.py
josericardo: Ok @pablohoffman. #42 fixes the first concern. #41 fixes the second :)

shub/auth.py
```diff
@@ -0,0 +1,48 @@
...
+    if key:
+        return key
+
+
+def write_key_netrc(key):
+    descriptor = os.open(
+        NETRC_FILE,
+        os.O_CREAT | os.O_RDWR | os.O_APPEND, 0o600)
+    with os.fdopen(descriptor, 'a+') as out:
+        line = 'machine scrapinghub.com login {0} password ""\n'.format(key)
```
chekunkov: btw, looks like this is not valid .netrc entry and that confuses a lot. If you try the same approach for `storage.scrapinghub.com` (put key into login section and leave password with `""`) and use it with `curl -sn` or `httpie` you'll find out that `""` from password field are also encoded which produces incorrect basic auth token and leads to 401. 

what confuses me is that we are using .netrc not following it's spec thus I don't really understand why we use `.netrc` at all.

pablohoffman: `.netrc` has been dropped in favor of YAML configuration for everything:
https://github.com/scrapinghub/shub/wiki/Configuration

That is already implemented.

(netrc has actually been kept for backwards compat, but will soon be dropped completely)

jdemaeyer: > (netrc has actually been kept for backwards compat, but will soon be dropped completely)

`.netrc` was already completely dropped. Only project-specific configuration is still read from `scrapy.cfg` (if there's no project `scrapinghub.yml` found)

chekunkov: > .netrc was already completely dropped.
@pablohoffman @jdemaeyer I remember last time when we dropped scrapy.cfg in favour of .netrc that was PITA for our clients and they didn't like the update process. I checked https://github.com/scrapinghub/shub/pull/79 and haven't found anything that looks like migration... shub can check ~/.scrapy.cfg and ~/.netrc and compose ~/.scrapinghub.yml based on these files. Can we add such a migration before releasing new version?

PR #41, opened 2015-06-26
'shub login' suggesting API key defined in scrapy.cfg

josericardo:
This should be merged after #40.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 click
 requests
-scrapy
 six
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 click
 requests
 scrapy
+six
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 click
 requests
+scrapy
--- a/shub/login.py
+++ b/shub/login.py
@@ -1,19 +1,15 @@
-import os, re, click
-from shub.utils import get_key_netrc, NETRC_FILE
+import re, click
+from shub import auth
 
 @click.command(help='add Scrapinghug API key into the netrc file')
 @click.pass_context
 def cli(context):
-    if get_key_netrc():
+    if auth.get_key_netrc():
         context.fail('Key already exists in netrc file')
     key = raw_input('Insert your Scrapinghub API key: ')
+
     if key and is_valid_key(key):
-        descriptor = os.open(
-            NETRC_FILE,
-            os.O_CREAT | os.O_RDWR | os.O_APPEND, 0o600)
-        with os.fdopen(descriptor, 'a+') as out:
-            line = 'machine scrapinghub.com login {0} password ""\n'.format(key)
-            out.write(line)
+        auth.write_key_netrc(key)
     else:
         context.fail('Invalid key')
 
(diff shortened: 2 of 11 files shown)


pablohoffman:
nice, tests!

josericardo:
Indeed, it turns out the login wasn't working in python 3. Just fixed.

requirements.txt
```diff
@@ -1,2 +1,4 @@
 click
 requests
+scrapy
+six
```
eliasdorneles: `six` should be added to `install_requires` in setup.py since it's a hard dependency now.

pablohoffman: Fine to add `six`. We shouldn't require `scrapy` as hard-dep though.

eliasdorneles: Agreed. But it's okay on requirements.txt, right?

pablohoffman: If `requirements.txt` is for development only, then yes.

eliasdorneles: :+1: that's how I see it, it represents the things we use in development and run our tests against.

and in setup.py we leave the dependencies as loosely defined as possible.

shub/login.py
```diff
@@ -1,21 +1,56 @@
-import os, re, click
-from shub.utils import get_key_netrc, NETRC_FILE
+import re
+import click
+import os
+from shub import auth
+from scrapy.utils.conf import get_config
```
eliasdorneles: talking about dependencies, this is making us dependant on Scrapy now for login.

eliasdorneles: I'm unsure about this now... Scrapy brings a lot of dependencies with it, but it's weird to require all that just for a few functionality.

@pablohoffman @josericardo would you be okay with duplicating the functions from scrapy.utils.conf here?

pablohoffman: Good catch @eliasdorneles!

Looking for `scrapy.cfg` file should only happen if scrapy is available, but otherwise it should still work (not support reading API keys from scrapy.cfg).

The tests should run without `scrapy` installed, otherwise it will be hard to keep this somewhat intricate logic working.

eliasdorneles: ah, yup, that's better than duplicating, +1.

I think just moving the import to the try block in `_find_api_key`.

josericardo: I considered copying scrapy.utils.conf when I noticed all the deps that were being loaded just for it. On the other hand, does it make sense to use shub if the user is not working on a scrapy project?

pablohoffman: Maybe not yet, but soon enough. For example, if you want to play with the data, not deploy anything.

```
$ shub login
$ shub items 123/45/67 | grep 'product title' | python -mjson.tool
```

pablohoffman: Or ...

```
shub dataset:download target.com
```

:wink:

eliasdorneles: yeah. =)
shub is for any Scrapinghub user -- not just the ones developing Scrapy projects.
I'd argue even in the current state it is already useful for people doing support. ;)

josericardo: Ok, updated. Scrapy is optional again (not for tests).

josericardo: Good to hear it's able to help more people than I thought :D

PR #42, opened 2015-06-26
Lower the precedence of $SHUB_APIKEY

josericardo:
Giving higher precedence to the .netrc API key will avoid confusion for
those who have $SHUB_APIKEY set and have also logged in with 'shub login'.

This should be merged after #41


pablohoffman:
doesn't this pull request need to be rebased?. It adds a `_prompt_for_key()` function that already exists on `shub.login`

josericardo:
Sure, done :)

PR #43, opened 2015-06-29
Make subcommands fail early if the user is not logged in

josericardo:
This avoids the execution of useless steps.

### DIFF
--- a/shub/deploy_egg.py
+++ b/shub/deploy_egg.py
@@ -8,6 +8,7 @@ import click
 from shub.click_utils import log, fail
 from shub import utils
 from shub.utils import run, decompress_egg_files
+from shub.auth import find_api_key
 
 
 @click.command(help="Build and deploy egg from source")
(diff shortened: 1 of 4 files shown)


josericardo:
It's probably a good idea to release a new minor version after this PR is merged @eliasdorneles @pablohoffman. We have fixed quite a few bugs since the last one.

pablohoffman:
I would do this but differently, why not do a initial check and define a list of commands that require being logged in, like we have a list of commands that require additional commands (`module_deps`)

josericardo:
I like it @pablohoffman. PR coming soon :)

eliasdorneles:
hm, I kinda liked this approach because it didn't need extra bookkeeping. :)

josericardo:
@eliasdorneles Yep and looking at the code the find_api_key calls still will have to be where they are. It looks like the code would get more fragile. I'm having trouble to see it as an improvement @pablohoffman (besides ensuring the login error will be displayed earlier).

josericardo:
Unless we do a large-ish refactoring to wrap all api requests behinds some kind of interface. Then it would be a good thing :)

pablohoffman:
So this fails with a python exception if no API key is defined?. How about a gentler error message asking the user to login?

josericardo:
We do: https://github.com/scrapinghub/shub/pull/37/files#diff-1e43fbb121ffe27f46a4e3a1cbe536b0R49. This exception is handled by click :)

pablohoffman:
all good then! :+1:

PR #45, opened 2015-07-09
Fix the "unknown" egg versions problem in deploy-reqs

### LINKED ISSUE #44: deploy-reqs is not finding some egg versions
We have some eggs deployed with ```unknown``` version.

Try:

-e git+https://github.com/scrapinghub/frontera@settings_override#egg=frontera

josericardo:
Reverts #21 and fixes #44 

Instead of using an alternate src dir, now we just ignore the folders
that don't look like egg dirs. The only know case for this is when we
try to install packages from git and the src dir is left in the same
folder as the eggs folders. So, it shouldn't be an issue.

### DIFF
--- a/shub/deploy_reqs.py
+++ b/shub/deploy_reqs.py
@@ -2,7 +2,6 @@ from __future__ import print_function
 import click
 import os
 import tempfile
-import shutil
 
 from shub.utils import run, decompress_egg_files
 from shub.click_utils import log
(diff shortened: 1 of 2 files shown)


pablohoffman:
can somebody review and merge this? @bbotella @eliasdorneles ?

bbotella:
Doesn't seem to be working in every situation. Checking with @josericardo for other approaches.

jdemaeyer:
Closing this since it seems to no longer be an issue, see #44

PR #47, opened 2015-07-10
Adds --setup-file parameter

bbotella:


eliasdorneles:
@bbotella could you describe what's the use case for this?

bbotella:
You may want to deploy a eggs to different dash projects in a different way, so your code can maintain different setup.py files. With this parameter, you can chose which setup.py file to use.

shub/deploy.py
```diff
@@ -140,17 +141,17 @@ def _upload_egg(target, eggpath, project, version, auth):
     return make_deploy_request(url, data, files, auth)
 
 
-def _build_egg():
+def _build_egg(setup_file):
     closest = closest_scrapy_cfg()
     os.chdir(os.path.dirname(closest))
-    if not os.path.exists('setup.py'):
+    if not os.path.exists(setup_file):
```
josericardo: I'm not sure this is OK for custom setup files. For example, if the user mistypes the setup file name it will silently generate a new setup file based on the template and (may) deploy without errors. 

I think it's a good idea to restrict this feature only when setup_file == 'setup.py'?

Thoughts? @bbotella @eliasdorneles ?

bbotella: And what about launching a generic warning message? Something like "somesetup.py doesn't exist. Creating it from template."

eliasdorneles: So, I think we should consider if we really want this feature.

I understand we needed it for an internal project with weird packaging and this change doesn't look like much, but the more arguments we add, the more decisions we are asking for our users to make.

IMO, it'd be better to fix the packaging of the internal project, keep shub's flow more straightforward and if some other user asks for this we consider adding it again.

josericardo: @bbotella from my experience users (myself included) don't pay too much attention to those msgs as long as there's a success in the end.

@eliasdorneles That's another option that I've considered. I checked the internal project and found out that we've been using two setup files for quite some time, so the use case seemed to be legitimate.

bbotella: Well, I think that adding options doesn't have to add complexity to the tool usage if we provide nice and logic defaults for that options. On the other hand, I agree with @eliasdorneles that this particular feature can be unnecessary if packages to be deploy are well planned and follow standards. BUT, if this feature doesn't affect the way users use the tool, and it doesn't break backwards compatibility, it provides a bit more of flexibility to the tool.
@josericardo That is right. Warning messages are usually ignored. So, the proposed behaviour is to throw an error if a setup file is given and it doesn't exist. Right?

eliasdorneles: > So, the proposed behaviour is to throw an error if a setup file is given and it doesn't exist.

+1, if we're keeping the feature, that's what I would expect.

josericardo: Yes, I still think that throwing an error is the way to go for user defined setup files.

josericardo:
@bbotella Any chance for you to take a look at this PR in the foreseeable future? :D

bbotella:
@josericardo Yeah, I'll add that error if file does not exist. :-)

eliasdorneles:
Sorry to be bothering, but do we still need the feature? :)

bbotella:
We don't "need" it anymore, but it is a halfdone feature that, in my opinion, doesn't make harm. If we prefer to keep it simple, let's just reject this PR.

eliasdorneles:
Every feature we add adds burden to maintainance.

Less is more. YAGNI.

The more options we add, the less predictable it is, the more combinations we need to test and the more complicated the interface is to use and the code to read.

This code was useful back then, now it's legacy code.
I vote for closing.

bbotella:
Ok, so, if @josericardo gives his ok I will close the PR. :-)

josericardo:
Closing is ok for me :)

PR #50, opened 2015-07-17
[MRG+1] Fix version names when setup.py --version returns many lines

josericardo:
This fixes the version names in the same way that
e6f0b1767b093791f457928d2d70bd32764fcf6a fixes the problem with egg
names in #49


eliasdorneles:
@josericardo think we could have tests for these?
as there are a lot of funny `setup.py` files out there, it will be nice if we collect samples starting from now. :)

josericardo:
Yes, it's a good idea, I just have to recall what was the problematic egg :D

PR #53, opened 2015-08-05
Stop returning an error for logged in login attempts

### LINKED ISSUE #46: login while already logged-in shows error
If one is already logged in, `shub login` throws an error, while it could say something less "dramatic"

Current behavior:

```
$ shub login
Usage: shub login [OPTIONS]

Error: Already logged in. To logout use: shub logout
```

- no need to show "Usage" as it's correct
- "Error" is misleading. it could say "You're already logged in. Nothing to do. If you want to login with another API key, use `logout` first", or something along those lines

josericardo:
Now we just advise the user to logout first if he wants to change the
credentials.

Fixes #46

### DIFF
--- a/shub/login.py
+++ b/shub/login.py
@@ -10,7 +10,8 @@ from six.moves import input
 @click.pass_context
 def cli(context):
     if auth.get_key_netrc():
-        context.fail('Already logged in. To logout use: shub logout')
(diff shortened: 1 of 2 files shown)


eliasdorneles:
Nice!

PR #54, opened 2015-08-06
Use the project id in the scrapy.cfg as default for deploy-reqs

### LINKED ISSUE #52: deploy-reqs should use the project_id defined in scrapy.cfg
deploy-reqs cli arguments are not consistent with other commands like deploy.
project_id is required but shoud be an optional argument and the command should call the same _get_project function that is defined in deploy.py

josericardo:
An alternative project id can still be specified via the option -p or
--project flags.

Fixes #52.

### DIFF
--- a/tests/test_deploy_reqs.py
+++ b/tests/test_deploy_reqs.py
@@ -6,23 +6,29 @@ from __future__ import print_function
 import unittest
 import os
 import tempfile
-from mock import Mock
+from mock import Mock, patch
+from click.testing import CliRunner
 
 from shub import deploy_reqs
 
 
 class TestDeployReqs(unittest.TestCase):
+    def setUp(self):
+        self.runner = CliRunner()
+        os.environ['SHUB_APIKEY'] = '1234'
+
     def test_can_decompress_downloaded_packages_and_call_deploy_reqs(self):
         # GIVEN
         requirements_file = self._write_tmp_requirements_file()
+        with self.runner.isolated_filesystem():
+            # WHEN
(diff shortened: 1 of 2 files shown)


pablohoffman:
LGTM, @eliasdorneles to merge

tests/test_deploy_reqs.py
```diff
@@ -6,23 +6,42 @@
...
 
 
+@patch('shub.deploy_reqs.utils.build_and_deploy_egg')
 class TestDeployReqs(unittest.TestCase):
-    def test_can_decompress_downloaded_packages_and_call_deploy_reqs(self):
+    VALID_APIKEY = '1234'
+
+    def setUp(self):
+        self.runner = CliRunner()
+        os.environ['SHUB_APIKEY'] = self.VALID_APIKEY
```
eliasdorneles: Hm, I'm not keen on this line -- side effects for the tests at process level.
It will be better with a context manager that restores the original value or unsets it, appropriately.

eliasdorneles: You can probably copy `os.environ` entirely and replace it inside the context manager.

All the rest LGTM! :+1:

josericardo: @eliasdorneles ok, CliRunner also allows us to override environment variables in an isolated way. Using it now. I'll refactor some tests after these PRs are merged to DRY them a little bit. This is repeated in other tests also.

eliasdorneles: :+1:

PR #55, opened 2015-08-06
Use requirements.txt as default for deploy_reqs

josericardo:
As that's usually the case. Alternative files can be defined via -r or
--requirements-file parameter

This should be merged after #54

### DIFF
--- a/shub/deploy_reqs.py
+++ b/shub/deploy_reqs.py
@@ -7,18 +7,22 @@ import shutil
 from shub.utils import run, decompress_egg_files
 from shub.click_utils import log
 from shub import utils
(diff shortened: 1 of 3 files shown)


jdemaeyer:
@josericardo can you rebase?

Was there a decision regarding
[> We need to re-do it in a backward compatible way.](https://github.com/scrapinghub/shub/issues/52#issuecomment-139619285)
? (/cc @eliasdorneles)

josericardo:
@jdemaeyer The problem with this PR is that I replace the positional args for flags. We still have to support the positional args.

jdemaeyer:
Superseded by #79

PR #59, opened 2015-09-03
[MRG+1] Revert "Use the project id in the scrapy.cfg as default for deploy-reqs"

josericardo:
Reverts scrapinghub/shub#54

This breaks the backwards compatibility of the command interface and we want to make a release today.


eliasdorneles:
+1, better to do this with more patience later, in a backwards compatible manner.

PR #62, opened 2015-11-04
Refactor the login tests

josericardo:
I felt the tests intents were lost among the mock machinery.

### DIFF
--- a/tests/test_login.py
+++ b/tests/test_login.py
@@ -5,12 +5,10 @@ from mock import Mock, patch
 from click.testing import CliRunner
 from shub import login
 
+VALID_KEY = 32 * '1'
 
 class LoginTest(unittest.TestCase):
-    VALID_KEY = 32 * '1'
-
     def setUp(self):
-        login.auth.NETRC_FILE = ''
         self.runner = CliRunner()
 
     def test_login_writes_input_key_to_netrc_file(self):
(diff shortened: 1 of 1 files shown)


josericardo:
@eliasdorneles how can I improve it?

tests/test_login.py
```diff
@@ -19,63 +17,63 @@ def test_login_writes_input_key_to_netrc_file(self):
...
+            for path, content in (files or {}).iteritems():
+                with open(path, 'w') as f:
+                    f.write(content)
 
-    def test_login_attempt_after_login_doesnt_lead_to_an_error(self):
-        with self.runner.isolated_filesystem() as fs:
+        def invoke():
+            return self.runner.invoke(login.cli, input=user_input, env=env)
+
+        def run():
```
josericardo: I know, this is smart method. Is it worth it?

eliasdorneles: I'm not sure how I feel about this.
I kinda like the reading of the other tests, but this feels a bit like sweeping the dirty under the rug. :)

josericardo: The problem is that the login command is just code that reads/writes files, input and makes network requests basically, there isn't much logic. 

We could try a different design for the login, more OO, for example. But I'm not sure it's worth the trouble right now as the feature is already pretty stable.

At least in the refactored code the test purpose is more obvious, IMO.

pablohoffman:
how do make a final call on this @josericardo @eliasdorneles?. do we need a third reviewer?

eliasdorneles:
No need, we've discussed it here and we're merging. :+1:

PR #66, opened 2015-12-03
added schedule and items commands

stummjr:
The `shub schedule` and `shub items` commands were added. Here is how they work:

#### shub schedule
Schedule a spider for execution. Example:
```
$ #scheduling 'my_spider' for project 26736
$ shub schedule my_spider -p 26736

$ #scheduling 'my_spyder' for project defined in scrapy.cfg
$ shub schedule my_spider

$ #passing arguments to the spider
$ shub schedule my_spider -p 26736 -a arg1=foo arg2=bar
```

#### shub items
Fetch items from a given job. Example:
```
$ #getting the items scraped by job 26736/2/71
$ shub items 26736/2/71
```

### DIFF
--- a/shub/schedule.py
+++ b/shub/schedule.py
@@ -1,0 +1,45 @@
+from ConfigParser import NoSectionError, NoOptionError
+import click
+from click import ClickException
+from hubstorage import HubstorageClient
+from shub import scrapycfg
+from shub.utils import find_api_key
+
+
+@click.command(help='Schedule a spider to run on Scrapy Cloud')
+@click.pass_context
+@click.argument("spider", type=click.STRING)
+@click.option("-p", "--project-id", help="the project ID", type=click.INT)
+@click.option("-a", "--argument", help="argument for the spider", multiple=True)
+def cli(context, project_id, spider, argument):
+    apikey = find_api_key()
+    if not apikey:
+        raise ClickException('Scrapinghub API key not found: please, run \'scrapy login\' first')
+    project_id = project_id or get_project_id_from_config()
+    job_key = schedule_spider(apikey, project_id, spider, argument)
+    click.echo(
+        'Spider {} scheduled: {}\nCheck the spider execution at: '
+        'https://dash.scrapinghub.com/p/{}/job/{}/{}'.format(spider, job_key, *job_key.split('/'))
+    )
+
+
+def schedule_spider(apikey, project_id, spider, arguments=()):
+    hc = HubstorageClient(auth=apikey)
+    project = hc.get_project(project_id)
+    if project.ids.spider(spider) is None:
+        raise ClickException(
+            'Spider \'{}\' doesn\'t exist in project {}'.format(
+                spider, project_id
+            )
+        )
+    job = hc.push_job(projectid=project_id, spidername=spider,
+                      spider_args=dict(x.split('=') for x in arguments))
+    return job.key
+
+
+def get_project_id_from_config():
+    try:
+        conf = scrapycfg.get_config()
+        return conf.get('deploy', 'project')
+    except (NoSectionError, NoOptionError):
+        raise ClickException('missing project id')
--- a/shub/schedule.py
+++ b/shub/schedule.py
@@ -1,7 +1,7 @@
 from ConfigParser import NoSectionError, NoOptionError
 import click
 from click import ClickException
-from hubstorage import HubstorageClient
+from scrapinghub import Connection
 from shub import scrapycfg
 from shub.utils import find_api_key
 
@@ -10,7 +10,7 @@ from shub.utils import find_api_key
 @click.pass_context
 @click.argument("spider", type=click.STRING)
 @click.option("-p", "--project-id", help="the project ID", type=click.INT)
-@click.option("-a", "--argument", help="argument for the spider", multiple=True)
+@click.option("-a", "--argument", help="argument for the spider (-a name=value)", multiple=True)
 def cli(context, project_id, spider, argument):
     apikey = find_api_key()
     if not apikey:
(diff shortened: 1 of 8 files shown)


shub/tool.py
```diff
@@ -24,6 +24,8 @@ def cli():
     "deploy_reqs": [],
     "logout": [],
     "version": [],
+    "items": ["hubstorage"],
```
pablohoffman: hubstorage should be a required dependency, and shipped with shub tool.

These commands should always work, not like deploy that requires scrapy to be present.

shub/schedule.py
```diff
@@ -0,0 +1,45 @@
+from ConfigParser import NoSectionError, NoOptionError
```
bertinatto: In Python 3 ConfigParser has been renamed to configparser.

bertinatto: I'd recommend using `configparser` from `six`.

tests/test_items.py
```diff
@@ -0,0 +1,26 @@
+import mock
+import unittest
+from click import ClickException
+from shub import items
+
+
+class ItemsTest(unittest.TestCase):
```
josericardo: It would be a little better if these tests used click.testing.CliRunner. It's more end-to-end (more coverage).

It wouldn't change much and we'd be sure the interfaces are running fine :)

shub/schedule.py
```diff
@@ -0,0 +1,41 @@
...
+import click
+from click import ClickException
+from scrapinghub import Connection
+from shub import scrapycfg
+from shub.utils import find_api_key
+
+
+@click.command(help='Schedule a spider to run on Scrapy Cloud')
+@click.pass_context
+@click.argument("spider", type=click.STRING)
```
bertinatto: Just a nitpick, but I think we should be consistent in the use of single or double quotes.

stummjr: :+1:

shub/schedule.py
```diff
@@ -0,0 +1,41 @@
...
+        raise ClickException('Scrapinghub API key not found: please, run \'scrapy login\' first')
+    project_id = project_id or get_project_id_from_config()
+    job_key = schedule_spider(apikey, project_id, spider, argument)
+    click.echo(
+        'Spider {} scheduled: {}\nCheck the spider execution at: '
+        'https://dash.scrapinghub.com/p/{}/job/{}/{}'.format(spider, job_key, *job_key.split('/'))
+    )
+
+
+def schedule_spider(apikey, project_id, spider, arguments=()):
```
josericardo: This method made think about "It's better to ask forgiveness than permission". Although I'm not familiar to the scrapinghub lib, can't you just try to schedule and handle the exceptions?

like simply calling `conn[project].schedule`. What do you think?

shub/schedule.py
bertinatto: When our dependencies add py3 support we might do the same for `shub`; so I think we should keep new code py3-compliant.

tests/test_items.py
```diff
@@ -0,0 +1,49 @@
...
+    @mock.patch('shub.items.find_api_key', autospec=True)
+    @mock.patch('shub.items.fetch_items_for_job', autospec=True)
+    def test_fetches_items_if_input_is_ok(self, mock_find_apikey, mock_fetch_items_for_job):
+        mock_find_apikey.return_value = 1
+        self.runner.invoke(items.cli, ['1/1/1'])
+        self.assertTrue(mock_fetch_items_for_job.called)
+
+    @mock.patch('shub.items.HubstorageClient', spec=True)
+    def test_fetch_items_for_job_with_invalid_job(self, mock_hs):
+        mock_hs.return_value.get_job.return_value.metadata = None
```
josericardo: Have we gone too far on this mocks? IMO, if we only mock `get_job` (returning a job) it would be more clear what the test is actually testing.

shub/schedule.py
stummjr: Sure we can! :+1: 
Way better. :-)

shub/schedule.py
josericardo: +1 @bertinatto

shub/schedule.py
eliasdorneles: I believe the point of raising ClickException here is sending a nicer message to user.

The messages can be improved -- instead of `"doesn't exist"`, it's better `"project {} not found"` (it can be that the project exists but the used credentials don't have access to it).

shub/schedule.py
```diff
@@ -0,0 +1,41 @@
+from ConfigParser import NoSectionError, NoOptionError
+import click
+from click import ClickException
+from scrapinghub import Connection
+from shub import scrapycfg
```
eliasdorneles: We could get something nice making this import lazy (moving it to inside `get_project_id_from_config()`): the command would not need Scrapy when you use `-p` argument.
If it doesn't complicate anything else, would be cool to have it. :)

shub/schedule.py
eliasdorneles: wow, that's a real nitpick. :D
I stopped worrying about single quotes vs double quotes when I stopped doing Perl and PHP. :)

eliasdorneles:
Nice work, @stummjr !
I added a bit to the nitpicking, but this looks good, well done! :)

shub/schedule.py
josericardo: As we're shipping shub as a binary now, does it still make sense to be picky about having scrapy or not? Aren't we always bundling it already?

tests/test_items.py
stummjr: :+1:

shub/schedule.py
josericardo: Oh, I suggested a refactoring. No changes on behavior :)

eliasdorneles: ah, sorry :)

josericardo:
Thanks @stummjr, well done indeed! ;D

bertinatto:
+1

stummjr:
Thank *you* guys for the help!

PR #67, opened 2015-12-04
Make the schedule command not raise exception when scrapy isn't installed

josericardo:


josericardo:
shub doesn't depend on scrapy anymore

PR #69, opened 2015-12-07
replace https with http in homepage url

pawelmhm:
because https does not work and value of "url" is displayed in pypi https://pypi.python.org/pypi/shub


eliasdorneles:
Thanks @pawelmhm !

PR #74, opened 2015-12-08
Fix utils test

bertinatto:


tests/test_utils.py
```diff
@@ -31,6 +32,11 @@ def check(cmd):
                 # then
                 self.assertEquals('lxml-3.4.4', version)
 
+    def test_raises_click_exception_with_invalid_job_id(self):
+        with self.assertRaises(ClickException):
```
eliasdorneles: since ClickException is quite generic, it's better to use `self.assertRaisesRegexp` and match on the exception message too.

bertinatto: :+1:

josericardo:
Thanks @bertinatto :)

PR #77, opened 2015-12-09
Unify and test fetching job resources

jdemaeyer:
Solves #72

`shub.items`, `shub.log` and `shub.requests` are very similar. This PR removes most of the code duplication by exporting the "Get this job from Hubstorage" functionality into a utility function.

Our use case (all commands need a job id as argument, perform checks, then access a job from HS) pretty much matches Click's [groups](http://click.pocoo.org/5/quickstart/#nesting-commands), we should consider implementing a "fetch from HS" group.


eliasdorneles:
:+1:

PR #80, opened 2015-12-15
Adding a tox target to freeze shub command line tool

jsargiot:
This new target will generate a binary of the platform on which is running (windows, linux or mac).

## Tox target

    $ tox -e freeze

The new target runs two commands:

* `pyinstaller ...`: creates the binary for the platform; and
* `.../shub version`: Runs the newly built binary, if there is a missing library it will fail. Since all commands are loaded on startup, if this command is successful means that all commands (and their dependencies) were loaded ok.

### Hooks

Two hooks were added to ensure that `hubstorage`and `shub` included all files and runtime dependencies.

### Runner

Added a shubrunner.py script to execute the cli function of the shub.tool module because pyinstaller doesn't support that call.


josericardo:
Hey @jsargiot :)

Just checking, would it be possible to add a test for this? Something like extracting the pyinstaller command to a script, running it from a test and checking the package was successfully built?

Overkill?

pablohoffman:
I asked @jsargiot to add it to travis, so its runs on every build and we confirm it works.

jsargiot:
> Just checking, would it be possible to add a test for this? Something like extracting the pyinstaller command to a script, running it from a test and checking the package was successfully built?

I've added a second command to the `freeze` target to run the `version` command. Since all commands are loaded on startup with their dependencies I can confirm that if it runs, all deps are ok. I'm not testing functionality, it's being tested in the default target

eliasdorneles:
hey, this is neat! :)

josericardo:
Thanks @jsargiot :D

PR #87, opened 2015-12-19
commands items/log/requests outputting JSON lines

### LINKED ISSUE #86: Let `shub items` outputs in JSON-lines format.
At this very moment `shub items` outputs results in a format of newline-separated stringified Python `dict`:

```
{u'name': u'item_0', u'foo': u'bar'}
{u'name': u'item_1', u'foo': u'bar'}
```

Personally I think it's better to make it JSON-lines:

```
{"name": "item_0", "foo": "bar"}
{"name": "item_1", "foo": "bar"}
```

Or if we need to keep its backward compatibility, I'd suggest at least we introduce something like `shub items --format=jsonl`.

stummjr:
This PR fixes #86 .

### DIFF
--- a/shub/log.py
+++ b/shub/log.py
@@ -1,6 +1,9 @@
-import click
+import logging
+from datetime import datetime
 
 from shub.utils import get_job
+
+import click
 
 
 @click.command(help='Get log of a given job on Scrapy Cloud')
@@ -8,4 +11,10 @@ from shub.utils import get_job
 def cli(job_id):
     job = get_job(job_id)
     for item in job.logs.iter_values():
-        click.echo(item)
+        click.echo(
+            "{} {} {}".format(
+                datetime.utcfromtimestamp(item['time']/1000),
+                logging.getLevelName(int(item['level'])),
+                item['message']
+            )
+        )
--- a/shub/items.py
+++ b/shub/items.py
@@ -7,5 +7,5 @@ from shub.utils import get_job
 @click.argument('job_id')
 def cli(job_id):
     job = get_job(job_id)
-    for item in job.items.iter_values():
+    for item in job.items.iter_json():
         click.echo(item)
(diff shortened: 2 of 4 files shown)


pablohoffman:
1. Shouldn't we actually test that the output of the commands is in the expected format?
2. I think text is a better default format for `shub log`, since you typically want to review logs manually to understand what happened (as with Scrapy logs) not programmatically process them. About requests, I'm not sure what's the best default, but raw json is probably not either. It'd depend on what's the typical use case for using `shub requests` which I'm not sure about just yet. It looks like requests was mostly added for completeness without much thinking about use cases.And that's fine, we can keep using json there until something better comes up.

josericardo:
+1 on `shub log` output to be text :)

jdemaeyer:
I agree on `shub log`.

The problem with testing whether the output is actually json lines formatted is that we would have to dig quite deep into python-hubstorage to plant our test data. And even then, python-hubstorage never really holds or jsonifies the data but merely [forwards the hubstorage response](https://github.com/scrapinghub/python-hubstorage/blob/master/hubstorage/resourcetype.py#L29).

I think it's fine to simply make sure that our implementation uses `iter_json` here (which the test does), and rely on hubstorage/python-hubstorage to work as advertised.

pablohoffman:
OK, I thought we were going to mock hubstorage anyway.

shub/log.py
```diff
@@ -1,11 +1,11 @@
...
 
+import click
+
 
 @click.command(help='Get log of a given job on Scrapy Cloud')
 @click.argument('job_id')
 def cli(job_id):
     job = get_job(job_id)
-    for item in job.logs.iter_values():
+    for item in job.logs.iter_json(format='text'):
```
dangra: this will fail at first retried error at [hubstorage/resourcetype.py#L60](https://github.com/scrapinghub/python-hubstorage/blob/master/hubstorage/resourcetype.py#L60) called from  [hubstorage/resourcetype.py#L81](https://github.com/scrapinghub/python-hubstorage/blob/master/hubstorage/resourcetype.py#L81)

jdemaeyer:
Looks great :) could you rebase?

stummjr:
:+1:

PR #88, opened 2015-12-21
Allow scrapinghub.yml targets in items/log/requests

jdemaeyer:
Adds support for
 * `shub items 1/1` (fetches from default project)
 * `shub items otherproject/1/1` (fetches from `otherproject` defined in `scrapinghub.yml`)

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -183,18 +183,31 @@ def _get_egg_info(name):
     return (egg_filename, egg_path)
 
 
-def validate_jobid(jobid):
-    if not bool(re.match(r'\d+/\d+/\d+$', jobid)):
-        err = 'Job ID {} is invalid. Format should be\
-               projectid/spiderid/jobid'.format(jobid)
-        raise ClickException(err)
+def get_job_specs(job):
+    """
+    Parse job identifier into valid job id and corresponding API key.
 
-
-def get_job(jobid):
-    # XXX: Lazy import to avoid circular dependency. Needs refactoring
+    With projects default=10 and external=20 defined in config:
+    * 1/1 -> 10/1/1
+    * 2/2/2 -> 2/2/2
+    * external/2/2 -> 20/2/2
+    """
+    match = re.match(r'^((\w+)/)?(\d+/\d+)$', job)
+    if not match:
+        raise ClickException(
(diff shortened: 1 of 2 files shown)


shub/utils.py
```diff
@@ -183,18 +183,31 @@ def _get_egg_info(name):
...
-def get_job(jobid):
-    # XXX: Lazy import to avoid circular dependency. Needs refactoring
+    With projects default=10 and external=20 defined in config:
+    * 1/1 -> 10/1/1
+    * 2/2/2 -> 2/2/2
+    * external/2/2 -> 20/2/2
+    """
+    match = re.match(r'^((\w+)/)?(\d+/\d+)$', job)
+    if not match:
+        raise ClickException(
```
josericardo: Would it be a good idea to raise a new exception (that inherits from ClickException) here?

tests/test_utils.py
```diff
@@ -37,13 +37,26 @@ def check(cmd):
...
+        _test_specs('10/20/30', '10/20/30', 'default')
+        _test_specs('2/3', '1/2/3', 'default')
+        _test_specs('default/2/3', '1/2/3', 'default')
+        _test_specs('prod/2/3', '2/2/3', 'default')
+        _test_specs('vagrant/2/3', '3/2/3', 'vagrant')
+
+    def test_get_job_specs_validates_jobid(self):
+        invalid_job_ids = ['/1/1', '123', '1/2/a', '1//']
         for job_id in invalid_job_ids:
             self.assertRaisesRegexp(
```
josericardo: If a custom exception is raised, we can just catch it here, without caring abou the msg. This would lead to a more flexible test.

shub/utils.py
jdemaeyer: Indeed. There are a couple of tests that check for specific strings in `result.output` (after `result = runner.invoke(...)`) instead of the more sensible `result.exception`. As a bonus, we can set the exit code per exception type (and even default messages if we want), allowing us to drop [`click_utils`](https://github.com/scrapinghub/shub/blob/master/shub/click_utils.py) (which isn't used consistently anyways right now)

So +1 :)

jdemaeyer:
Merge this and add custom exceptions consistently throughout shub in another PR?

PR #89, opened 2015-12-21
Make sure API keys are always returned as string

jdemaeyer:
Without this, API keys that (by chance) contain only digits will be parsed as integers by PyYAML and run into trouble [here](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub.py#L64)


pablohoffman:
ah, the perks of YAML! :)

dangra:
LGTM :)

PR #103, opened 2015-12-31
Migrations for new configuration

jdemaeyer:
Auto-migration for `.netrc` and `.scrapy.cfg` into `.scrapinghub.yml` and for `scrapy.cfg` into `scrapinghub.yml`.

I think this is a solution that minimises the pain we induce users during migration

### DIFF
--- a/shub/config.py
+++ b/shub/config.py
@@ -1,6 +1,6 @@
-from ConfigParser import SafeConfigParser
 import contextlib
-import os.path
+import netrc
+import os
 import time
 
 import click
@@ -10,10 +10,12 @@ import ruamel.yaml as yaml
 from shub.exceptions import (BadParameterException, BadConfigException,
                              ConfigParseException, MissingAuthException,
                              NotFoundException)
-from shub.utils import closest_file, pwd_hg_version, pwd_git_version
+from shub.utils import (closest_file, get_scrapycfg_targets, get_sources,
+                        pwd_hg_version, pwd_git_version)
 
 
 GLOBAL_SCRAPINGHUB_YML_PATH = os.path.expanduser("~/.scrapinghub.yml")
+NETRC_PATH = os.path.expanduser('~/_netrc' if os.name == 'nt' else '~/.netrc')
 
 
 class ShubConfig(object):
@@ -32,6 +34,8 @@ class ShubConfig(object):
         """Load Scrapinghub configuration from stream."""
         try:
             yaml_cfg = yaml.safe_load(stream)
+            if not yaml_cfg:
+                return
             for option in ('projects', 'endpoints', 'apikeys'):
                 getattr(self, option).update(yaml_cfg.get(option, {}))
             self.version = yaml_cfg.get('version', self.version)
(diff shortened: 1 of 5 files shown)


shub/config.py
```diff
@@ -130,42 +160,76 @@ def get_target(self, target, auth_required=True):
...
+done in a file called '.scrapinghub.yml' in your home directory.
+
+But no worries, shub has automatically migrated your global settings to
+~/.scrapinghub.yml. For migrating project settings, simply run
+'shub migrate-config' in the project directory.
+
+Visit http://doc.scrapinghub.com/shub.html for more information on the new
+configuration format and its benefits.
+
+Happy shubbing!
```
pablohoffman: Happy scraping!

pablohoffman:
+1 to the shub 1.6 "welcome message" 
+1 to migrate automatically without any wizard

-1 to keep supporting scrapy.cfg
-1 for the new `migrate-config` command

My proposal instead is to migrate everything at once (ie. run the migrate-config implicitly), displaying the welcome message only once (ie. when migrating the auth/home config).

shub/config.py
chekunkov: +1

chekunkov:
> My proposal instead is to migrate everything at once (ie. run the migrate-config implicitly), displaying the welcome message only once (ie. when migrating the auth/home config) and a single line afterwards (sent to stdout) like "Created scrapinghub.yml from scrapy.cfg"

no objections

chekunkov:
> -1 to keep supporting scrapy.cfg

+0. I'm afraid of possible conversion issues, exceptions can happen during conversion leading to incomplete or missing .yml config. There are 2 options:

- keep scrapy.cfg support for a couple of months, in this case even if something bad happened during conversion users won't have to install older versions and will be able to use old configs
- remove scrapy.cfg support - in this case exceptions should be handled appropriately and informative error message should be shown guiding user to either report a bug, try to figure out what happened and convert config manually or install older version.

jdemaeyer:
I'm fine with also auto-migrating the project `scrapy.cfg`.

Also, doing auto-migration and 'keep support for `scrapy.cfg`' are the same thing, what else are we gonna auto-migrate from? ;) It's just a question of how often we read `scrapy.cfg`.

I'll add instructions on how to get help and a fallback mechanism for when the conversion fails (I don't see how it could, but of course there's always the one case you don't think of)

pablohoffman:
I don't see how the conversion could fail either, since it's a pretty simple conversion of a few fields.

jdemaeyer:
done :) will squash after review

jdemaeyer:
Squashed =)

shub/exceptions.py
```diff
@@ -35,6 +35,11 @@ class NotFoundException(ShubException):
     default_msg = "Not found."
 
 
+class AlreadyExistsException(ShubException):
```
pablohoffman: Where is this exception used?

jdemaeyer: damn, thanks. That was a leftover from the `migrate-config` command.

PR #104, opened 2015-12-31
Use more meaningful exit codes

### LINKED ISSUE #99: Use more meaningful exit codes
How about replacing the generic exit code 1 we use everywhere right now with more meaningful exit codes (e.g. after [the `sysexits.h` convention](https://www.freebsd.org/cgi/man.cgi?query=sysexits&sektion=3)).

This would be super-easy after #97 and will make it easier for shell scripts to do error diagnostics when automating `shub` usage. The downside is that it's slightly backwards incompatible with scripts that check `exit_code == 1` instead of the more sane `exit_code != 0`

jdemaeyer:
Fixes #99 

This is still up for discussion, but if we do it I'd like to include it in the 1.6 release b/c that's the one where we change parts of the public API anyways.

### DIFF
--- a/shub/exceptions.py
+++ b/shub/exceptions.py
@@ -1,3 +1,10 @@
+"""
+shub-specific exceptions.
+
+Exit codes follow the sysexits.h convention:
+https://www.freebsd.org/cgi/man.cgi?query=sysexits&sektion=3
+"""
+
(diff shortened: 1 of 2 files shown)


pablohoffman:
Fine by me, although I see this more as a luxury than a need :)

PR #107, opened 2016-01-05
Adding OS X shub builder

jsargiot:
This PR adds OS X in the Travis build to run the build.

Since Travis doesn't support python in OS X I've added python installation with brew in `travis-workarounds.sh` (only when $TRAVIS_OS_NAME is `osx`)

The final build matrix is:
- TOXENV=py27,freeze in linux (Creates linux binary)
- TOXENV=py27,freeze in osx (Creates osx binary)

Link to working build: https://travis-ci.org/scrapinghub/shub/builds/100935289

@dangra @pablohoffman should also add TOXENV=py27 for osx?

### DIFF
--- a/.travis-workarounds.sh
+++ b/.travis-workarounds.sh
@@ -11,5 +11,18 @@ if [[ "${TOXENV}" == "pypy" ]]; then
     sudo rm -rf /usr/local/pypy/bin
 fi
 
+if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    brew update > /dev/null
+
+    brew install python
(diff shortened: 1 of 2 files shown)


dangra:
IMO we parametrise only on operating system and remove TOXENV from matrix. The testsuite is small so should be fast to run multiple envs in one build, but also we are sure artifacts are not build on test failures. Use a fixed env list like `tox -e py27,freeze`.

jsargiot:
@daniel: done

dangra:
LGTM.  + @pablohoffman @jdemaeyer

PR #105, opened 2016-01-02
Extend README

jdemaeyer:
Once we're happy with this I'll submit a PR to update the [doc page on scrapinghub.com](http://doc.scrapinghub.com/shub.html) (leaving out the advanged config section) and edit [this knowledge base entry](http://support.scrapinghub.com/topic/205471-deploying-your-project-to-scrapy-cloud/) (to the quickstart section?)

PR #109, opened 2016-01-06
Stop using scrapyd as API endpoint

jdemaeyer:
This is purely cosmetic, but `shub` can do more than deploy projects now, and the configured API endpoint should reflect that.

README should be updated when/if this is merged, it's not in here yet b/c it would conflict with #105

### DIFF
--- a/shub/config.py
+++ b/shub/config.py
@@ -20,7 +20,7 @@ NETRC_PATH = os.path.expanduser('~/_netrc' if os.name == 'nt' else '~/.netrc')
 
 class ShubConfig(object):
 
-    DEFAULT_ENDPOINT = 'https://dash.scrapinghub.com/api/scrapyd/'
+    DEFAULT_ENDPOINT = 'https://dash.scrapinghub.com/api/'
 
     def __init__(self):
         self.projects = {}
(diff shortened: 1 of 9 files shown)


pablohoffman:
thoughts on this one @dangra?

jdemaeyer:
Just saw it would conflict with some of @pawelmhm's use cases (ScrapyRT I guess?) in #112

pablohoffman:
FWIW, `shub` does not ensure future compatibility with scrapyd api. It may break at any time when we decide to add features / make changes to server side API. This was one of the main reasons for the original `scrapyd-client` fork.

pawelmhm:
I dont think there is conflict here @jdemaeyer configuration allows us to specify endpoint as we wish so all good.

dangra:
It makes complete sense, not objections from me. 

I keep thinking on backward compatibility plus a warning but normal users shouldn't set endpoints, only internal developers does afaik.

jdemaeyer:
> I dont think there is conflict here @jdemaeyer configuration allows us to specify endpoint as we wish so all good.

With this, `shub deploy` would automatically add `scrapyd/` after your configured target when deploying (since the real URL of `addversion.json` hasn't changed). You'd have to update the URLs of your ScrapyRT API endpoints (not in the shub config).

PR #110, opened 2016-01-06
Add --tail to shub log

josericardo:
Allows us to display only the last lines. I decided to add it after trying to tail a log of 80k lines :D

This was a quick hack, how can it be improved?

### DIFF
--- a/shub/log.py
+++ b/shub/log.py
@@ -36,10 +36,11 @@ SHORT_HELP = "Fetch log from Scrapy Cloud"
 @click.argument('job_id')
 @click.option('-f', '--follow', help='output new log entries as they are '
               'produced', is_flag=True)
-def cli(job_id, follow):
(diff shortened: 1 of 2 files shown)


pablohoffman:
You can always pipe to `tail -f`

josericardo:
Isn't that true? :facepalm:

jdemaeyer:
Hmm we could avoid to download the full 80k lines from the server if we got the number of lines from the server first (`job.logs.stats()['totals']['input_values']`), then start at `total_lines - tail_length` or something. Worth implementing? Or if not, maybe as default only download at most the latest 1,000 lines and have a flag to download the full log?

eliasdorneles:
@jdemaeyer I was thinking the same thing -- that'd be a nice feature. :)

josericardo:
Yeah, despite the noise, downloading 80k lines took some minutes.

+1 on a way to avoid that

PR #111, opened 2016-01-06
Make deploy work on projects that replaced scrapy.cfg for scrapinghub.yml

josericardo:
The idea is to allow projects with scrapy.cfg (as long as they have a scrapinghub.yml), right?


josericardo:
Uops again, scrapy.cfg can't be removed.

PR #117, opened 2016-01-11
Check for newer version using GitHub releases

### LINKED ISSUE #83: Check for updates
We will soon start distributing `shub` in a binary form, outside standard channels (like pypi or apt). So `shub` needs a way to check if there's a new update available for it.

I propose using Github releases API, and make sure we keep doing releases here. The check should be cached (maybe in `~/.scrapinghub.yml`) so as not to check on every invocation, but once a day (or something like that). It should also never fail (if there's no internet connection, for example) and cache failures as well.

jdemaeyer:
Fixes #83 

Once per day, fetch information on latest release through GitHub API (https://api.github.com/repos/scrapinghub/shub/releases/latest). On every shub run, compare it to the used shub version and show a message when there's a new version available. Never interferes with shub's functionality.

The latest release data is cached in `~.config/shub/` on Linux/OSX and in the `AppData`-thingy folder on Windows. Does not fail when the cache cannot be written or contains garbage.


shub/tool.py
```diff
@@ -23,7 +24,14 @@
 @click.group(help=HELP, short_help=SHORT_HELP, epilog=EPILOG)
 @click.version_option(shub.__version__)
 def cli():
-    pass
+    update_url = update_available()
+    if update_url:
+        click.echo(
```
jdemaeyer: Message ok?

pablohoffman: I'd keep it to one line:

    A new shub version is available. Update via pip or get it at http://github.com///

shub/utils.py
```diff
@@ -363,3 +366,59 @@ def job_resource_iter(job, iter_func, follow=True, key_func=None):
...
+
+
+def latest_github_release(force_update=False, timeout=1., cache=None):
+    """
+    Get GitHub data for latest shub release. If it was already requested today,
+    return a cached version unless ``force_update`` is set to ``True``.
+    """
+    REQ_URL = "https://api.github.com/repos/scrapinghub/shub/releases/latest"
+    cache = cache or os.path.join(click.get_app_dir('shub'),
+                                  'latest_release_cache.txt')
```
pablohoffman: `last_release.txt`?. just to keep the name shorter.

shub/utils.py
```diff
@@ -363,3 +366,59 @@ def job_resource_iter(job, iter_func, follow=True, key_func=None):
...
         time.sleep(15)
+
+
+def latest_github_release(force_update=False, timeout=1., cache=None):
+    """
+    Get GitHub data for latest shub release. If it was already requested today,
+    return a cached version unless ``force_update`` is set to ``True``.
+    """
+    REQ_URL = "https://api.github.com/repos/scrapinghub/shub/releases/latest"
+    cache = cache or os.path.join(click.get_app_dir('shub'),
```
pablohoffman: `scrapinghub` probably better for the app name

PR #130, opened 2016-01-14
Deploy onboarding + maintenance

### LINKED ISSUE #126: Deploy onboarding
It would be nice to add a small wizard when people use `shub deploy` for the first time, to ease the onboarding process. Then, all we need to say in Dash is use `shub deploy` and answer the questions.

```
$ shub deploy
Please login first with: shub login

$ shub login
...

$ shub deploy
Project ID you would like to deploy to: NNNNN
(...maybe validate access to the project? like shub login validates api key...)
Save project ID into scrapinghub.yml [Y/n]?
```

The questions will be asked if no `scrapinghub.yml` is found in the same dir as `scrapy.cfg`.
If you enter `Y` to the last question, the `scrapinghub.yml` file will be created with a default project set to the ID (NNNN) specified and you won't be asked again

jdemaeyer:
Add a wizard when users run `shub deploy` but have no `scrapinghub.yml` (and no deploy configuration in `scrapy.cfg`). Users will be prompted for a project ID, access rights to this project will be checked, and users can save it as default to `scrapinghub.yml` if wished.

Also contains some additional maintenance:
* PEP8ify `shub/deploy.py`
* Clean up `shub.deploy.cli()` (outsource `list_targets`)
* Rename `shub.config.update_config()` to `update_yaml_dict()`, since `update_config()` does not work on `ShubConfig` objects
* Update migration banner to reflect next shub version

Resolves #126.

### DIFF
--- a/shub/config.py
+++ b/shub/config.py
@@ -81,7 +81,7 @@ class ShubConfig(object):
             self._load_scrapycfg_target(tname, t)
 
     def save(self, path=None):
-        with update_config(path) as yml:
+        with update_yaml_dict(path) as yml:
             yml['projects'] = self.projects
             # Write "123" instead of "'123'"
             for target, project in yml['projects'].iteritems():
@@ -290,7 +290,7 @@ def load_shub_config(load_global=True, load_local=True, load_env=True):
 
 
 @contextlib.contextmanager
-def update_config(conf_path=None):
+def update_yaml_dict(conf_path=None):
     """
     Context manager for updating a YAML file while preserving key ordering and
     comments.
--- a/shub/config.py
+++ b/shub/config.py
@@ -182,7 +182,7 @@ class ShubConfig(object):
 
 MIGRATION_BANNER = """
 -------------------------------------------------------------------------------
-Welcome to shub v1.6!
+Welcome to shub version 2!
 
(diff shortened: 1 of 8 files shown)


shub/config.py
```diff
@@ -182,7 +182,7 @@ def get_target(self, target, auth_required=True):
 
 MIGRATION_BANNER = """
 -------------------------------------------------------------------------------
-Welcome to shub v1.6!
+Welcome to shub 2.0!
```
jsargiot: Maybe it's better to use shub.__version__, what do you think?

jdemaeyer: I was trying to avoid the impression that this is a welcome message to one specific version, whereas it is actually a welcome message to "shub configured through sh.yml" (e.g. v2+). But "2.0" is indeed not optimal either and could become confusing later. Maybe "v2"?

jsargiot: I get your point, "v2" does sounds better to me

pablohoffman: or "shub version 2"?

jdemaeyer: Also good, updated.

pablohoffman:
Looks good!. @josericardo wanna take a look before merging?

PR #131, opened 2016-01-14
Fix and clean up auto-versioning

### LINKED ISSUE #128: Fix broken auto-versioning
The ["get version control commit" commands](https://github.com/scrapinghub/shub/blob/master/shub/login.py#L55) throw an unhandled exception when their version control tool is not available.

This breaks deploying (shub exits with a traceback) when
- The user uses mercurial for version control and has not installed git
- The user uses bazaar for version control and has not installed both git and mercurial
- The user uses no version control and also has not installed all of the above tools

The exceptions should be either caught in the `pwd_git_version()` (and similar) util functions or in [`ShubConfig.get_version()`](https://github.com/scrapinghub/shub/blob/master/shub/config.py#L158).

See: https://stackoverflow.com/questions/377017/test-if-executable-exists-in-python

jdemaeyer:
Resolves #128

Also contains some maintenance (unification and cleanup)

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -11,11 +11,10 @@ import warnings
 
 from collections import deque
 from ConfigParser import SafeConfigParser
+from distutils.spawn import find_executable
 from distutils.version import StrictVersion
 from glob import glob
 from importlib import import_module
-from os import devnull
-from os.path import isdir
 from tempfile import NamedTemporaryFile
 from six.moves.urllib.parse import urljoin
 from subprocess import Popen, PIPE, CalledProcessError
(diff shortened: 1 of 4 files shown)


shub/utils.py
```diff
@@ -200,15 +237,12 @@ def _get_dependency_name():
...
-        return pwd_bzr_version()
-
-    version = _last_line_of(run('python setup.py --version'))
-    return "%s-%s" % (name, version)
+    # XXX: Why is the name included in the version? Keeping it that way in case
+    #      there was a good reason behind it
+    if any(os.path.isdir(dirname) for dirname in ('.git', '.hg', '.bzr')):
+        return pwd_version()
+    else:
+        return "%s-%s" % (name, pwd_version())
```
jdemaeyer: This returns strings like "lxml-3.4.4" as version, despite "lxml" already being save in the egg's 'name' attribute on Dash. [There's even a test that ensures this.](https://github.com/scrapinghub/shub/blob/master/tests/test_utils.py#L24) I've talked to @josericardo and he says this was copied from some old legacy deploy scripts. Any objections to dropping it? I.e. make it return "3.4.4" as version (with the name still saved as 'name' attribute)? (/cc @dangra @pablohoffman)

dangra: I see no reason to include name in version.

PR #133, opened 2016-01-15
Fix standalone deploy and improve behaviour for missing dependencies

jdemaeyer:
Stand-alone deploy did not work previously as `sys.executable` does not point to the python interpreter when frozen.

Also introduces error messages when `python` or `pip` are not available on the system but are required by a command.


pablohoffman:
good catch!

pablohoffman:
please rebase and let's merge

jdemaeyer:
done

PR #135, opened 2016-01-16
use params argument of the requests library

carlosp420:
When constructing an URL with key/value data, the requests library allows passing
them as a dictionary in the "params" argument.

I think that line is easier to read after this change.

See http://docs.python-requests.org/en/latest/user/quickstart/#passing-parameters-in-urls


dangra:
:+1:

dangra:
there is another line to fix like this in `shub/fetch_eggs.py`

carlosp420:
done!

PR #138, opened 2016-01-18
Use "with" statement to open files

carlosp420:
It is safer than using open and then close.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,14 +47,13 @@ This will save your Scrapinghub API key to a file in your home directory
 with your Scrapinghub account.
 
 Next, navigate to a Scrapy project that you wish to upload to Scrapinghub. You
-can deploy it to Scrapy Cloud by providing the Scrapinghub project ID, e.g.::
+can deploy it to Scrapy Cloud via::
 
-    shub deploy 12345
+    shub deploy
 
-Of course, it would be cumbersome if you had to re-enter the project ID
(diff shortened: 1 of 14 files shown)


carlosp420:
According to the AppVeyor logs, it is the installation of the requirement `rualmel.yaml` that fails. This has happened before: https://bitbucket.org/ruamel/yaml/issues/14/cant-install-ruamelyaml-package-under

carlosp420:
Yep, a new version of `rualmel.yaml` came out today https://pypi.python.org/pypi/ruamel.yaml
I will try to fall back to the previous version.

jdemaeyer:
Thx for the maintenance @carlosp420 

I suggest we wait for Anthon to fix it rather than force the old version

jdemaeyer:
Alternatively, if this keeps happening and annoys us too much, we can drop support for keeping user comments in yaml and go back to plain PyYAML.

jdemaeyer:
Should be good now.

carlosp420:
thanks for comments @jdemaeyer. Version 0.10.15 builds without errors now.

jdemaeyer:
great, can you rebase? :)

carlosp420:
well, I managed to screw up this pull request. Tried to fix it but made it worse. Closing and resubmitting as  #139
sorry :crying_cat_face:

PR #140, opened 2016-01-18
[WIP] Drop dependency to external commands

### LINKED ISSUE #129: Avoid dependencies on external tools
`shub` opens new Python sessions through the `subprocess` module. In particular, `shub deploy` depends on an available `python` executable (to build an egg through `python setup.py`), and `shub deploy-egg` and `shub deploy-reqs` additionally depend on `pip`.

While PyInstaller does bundle the Python interpreter, it is not an executable file (for Windows builds it bundles `python27.dll` and some additional libraries, which it then loads in the bootloader I guess). I have sent a question (yet to be released by a moderator) to the [PyInstaller mailing list](https://groups.google.com/forum/#!forum/pyinstaller) regarding how to call a new interpreter instance.

For `pip`, for now  we have settled in a Slack dicussion on deactivating the commands when `pip` is not available (for `deploy-egg` we should look into only deactivating the `--from-pypi` switch). There is a mailing list thread on bundling it [here](https://groups.google.com/forum/#!topic/pyinstaller/qZsciPL90c8).

jdemaeyer:
:boom: **DO NOT MERGE** :boom:

Resolves #129 

Makes `shub` Batteries Included™ by:
- Untarring/unzipping with the Python library instead of calling `unzip` or `tar` (which are seldomly available on Windows)
- Using `pip` as Python module instead of calling it as subprocess

Now if only I knew how to call a separate Python interpreter using PyInstaller...

This is WIP. ~~It works fine when I install `shub` via `pip` but makes problems in the stand-alone versions. In particular, `pip` fails to download any packages when I run it under Windows, for no apparent reason (but has no problems doing it when I run it under Linux using `wine`...).~~

Update: First running `pip.exe install -d tmp requests` (i.e. downloading `requests` with the system-wide `pip` install), then running `shub deploy-egg --from-pypi=requests` (which will read `requests` from the pypi cache) works. I think the problem is related to the frozen version not being able to validate PyPIs SSL certificate.

Update 2: Yup. Fixed it

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -94,7 +94,18 @@ def patch_sys_executable():
     """
     if getattr(sys, 'frozen', False):
         orig_exe = sys.executable
-        sys.executable = find_exe('python')
+        try:
+            py_exe = find_exe('python2')
+        except NotFoundException:
+            try:
+                # Will raise NotFoundException if no Python installation found
+                py_exe = find_exe('python')
+                if '2.' not in get_cmd_output([py_exe, '--version']):
+                    raise NotFoundException
+            except NotFoundException:
+                # Explicitly ask for installing 2.7
+                raise NotFoundException("Please install Python 2.7")
+        sys.executable = py_exe
         yield
         sys.executable = orig_exe
     else:
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -147,14 +147,14 @@ def pwd_version():
     if not ver:
         ver = pwd_bzr_version()
     if not ver and os.path.isfile('setup.py'):
-        ver = _last_line_of(run('python setup.py --version'))
+        ver = _last_line_of(run_python(['setup.py', '--version']))
     if not ver:
         closest_scrapycfg = closest_file('scrapy.cfg')
         if closest_scrapycfg:
             setuppy = os.path.join(os.path.dirname(closest_scrapycfg),
                                    'setup.py')
             if os.path.isfile(setuppy):
-                ver = _last_line_of(run('python {} --version'.format(setuppy)))
+                ver = _last_line_of(run_python([setuppy, '--version']))
     if not ver:
         ver = str(int(time.time()))
     return ver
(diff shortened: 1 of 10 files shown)


dangra:
do you think it is possible to add an integration test case for frozen binaries related to egg building?

pablohoffman:
:heart:

dangra:
lovely

shub/utils.py
```diff
@@ -80,6 +82,25 @@ def _is_deploy_successful(last_logs):
...
+    Context manager that monkey-patches sys.executable to point to the Python
+    interpreter.
+
+    Some scripts, in particular pip, depend on sys.executable pointing to the
+    Python interpreter. When frozen, however, sys.executable points to the
+    stand-alone file (i.e. the frozen script).
+    """
+    if getattr(sys, 'frozen', False):
+        orig_exe = sys.executable
+        sys.executable = find_exe('python')
```
dangra: this is still looking for external python binary no? any chance we can confirm it is python2 ?

jdemaeyer:
Alright, fixed the "doesn't download any packages in Windows" issue. `requests` couldn't find its `cacert.pem` file.

shub/utils.py
jdemaeyer: Indeed. It's a bit tricky because the Windows binary is always called `python.exe` (with no symlink from `python2`).

We could parse the output of `python --version` I guess.

Again this would be much nicer if we found a way to call a new Python instance from PyInstaller

jdemaeyer:
> do you think it is possible to add an integration test case for frozen binaries related to egg building?

Hmm a very good question. We won't be able to mock anything when we invoke the frozen binary. Maybe if we added a `--dry-run` flag or similar to `deploy-egg`.

dangra:
It is possible to run a webserver and use .scrapinghub.yml to change default endpoint. hijacking /etc/hosts on travis builds is only possible on non-docker builds, no idea if possible at all in appveyor.

jdemaeyer:
> It is possible to run a webserver and use .scrapinghub.yml to change default endpoint.

We could use this to improve our end-to-end tests, then have them run on both the package and stand-alone versions.

jdemaeyer:
OK I think this is somewhat complete. I have yet to play around with the binaries.

For the extensive end-to-end tests including binaries I think we should move to a new PR

tox.ini
```diff
@@ -17,6 +17,6 @@ deps =
     -rrequirements.txt
     pyinstaller
 commands =
-    pyinstaller --clean -y -F -n shub --additional-hooks-dir=./freeze/hooks --icon=./freeze/spider-down.ico ./freeze/shubrunner.py
+    pyinstaller --clean -y -F -n shub --additional-hooks-dir=./freeze/hooks --runtime-hook=./freeze/hooks/runtime-hooks.py --icon=./freeze/spider-down.ico ./freeze/shubrunner.py
     {toxinidir}/dist/shub version
```
dangra: is it too crazy if we include all test dependencies in the bundle and add a command like `shub run-tests` ?

PR #141, opened 2016-01-18
avoid nested if statements

carlosp420:


pablohoffman:
Logic is not preserved

carlosp420:
thanks for feedback Pablo. It took me a while to realize my error. I even had to write test cases to figure it out.
Maybe the test cases can be useful.

pablohoffman:
No problem. Sorry for the succinctness of the message, but I was on my phone.

jdemaeyer:
Test cases are always useful :)

I like splitting up this function, I'm tired of PyLint complaining about its complexity. Maybe instead of outsourcing the "If deploy successful" part we could outsource the whole "handle deploy response" part (i.e. the `with NamedTemporaryFile` block)? I think it makes more sense as a self-contained unit.

tests/test_utils.py
```diff
@@ -258,6 +263,30 @@ class MockException(Exception):
         with self.assertRaises(MockException):
             utils.update_available(silent_fail=False)
 
+    @patch('shub.utils._is_deploy_successful', return_value=True)
+    def test_echo_short_log_if_deployed_true(self, mock_dep_success):
+        log_file = LogFile()
```
jdemaeyer: You could also use a `mock.Mock` instance. Then you wouldn't have to define your own class

tests/test_utils.py
```diff
@@ -258,6 +263,30 @@ class MockException(Exception):
...
+        verbose = True
+        utils.echo_short_log_if_deployed(last_logs, log_file, verbose)
+        self.assertEqual(None, log_file.delete)
+
+        verbose = False
+        utils.echo_short_log_if_deployed(last_logs, log_file, verbose)
+        self.assertEqual(None, log_file.delete)
+
+    @patch('shub.utils._is_deploy_successful', return_value=None)
+    def test_echo_short_log_if_deployed_none(self, mock_dep_success):
```
jdemaeyer: These two tests have very similar overhead. Maybe there's a way to unify them? (You can change the return value during the test by setting `mock_dep_sucess.return_value`)

tests/test_utils.py
carlosp420: ok!

carlosp420:
Hi @jdemaeyer. I extracted the section you mentioned and added a test.

dangra:
@carlosp420 can you rebase?

carlosp420:
done! @dangra

dangra:
LGTM. Leaving the final call to @jdemaeyer

tests/test_utils.py
```diff
@@ -337,6 +337,37 @@ def _call(*args, **kwargs):
...
         self.assertIn('--no-binary=:all:', pipargs)
 
+    @patch('shub.utils._is_deploy_successful')
+    def test_echo_short_log_if_deployed(self, mock_dep_success):
+        log_file = Mock(delete=None)
+        last_logs = ["last log line"]
+
+        mock_dep_success.return_value = True
+        for verbose in [True, False]:
+            if verbose:
```
jdemaeyer: Hm, why are the `if` and `else` branches identical?

dangra: funny,  probably caused by rebase :)

carlosp420: doh! Fixed.

jdemaeyer:
Thanks @carlosp420

PR #143, opened 2016-01-20
Fix python2 enforcement in patch_sys_executable

jdemaeyer:
`python --version` print version to stderr, not stdout

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -100,7 +100,9 @@ def patch_sys_executable():
             try:
                 # Will raise NotFoundException if no Python installation found
                 py_exe = find_exe('python')
-                if '2.' not in get_cmd_output([py_exe, '--version']):
+                py_ver = subprocess.check_output([py_exe, '--version'],
+                                                 stderr=subprocess.STDOUT)
+                if '2.' not in py_ver:
                     raise NotFoundException
             except NotFoundException:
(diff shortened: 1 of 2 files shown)


shub/utils.py
```diff
@@ -100,7 +100,9 @@ def patch_sys_executable():
             try:
                 # Will raise NotFoundException if no Python installation found
                 py_exe = find_exe('python')
-                if '2.' not in get_cmd_output([py_exe, '--version']):
+                py_ver = subprocess.check_output([py_exe, '--version'],
+                                                 stderr=subprocess.STDOUT)
+                if '2.' not in py_ver:
```
dangra: I am tempted to be more strict and check for `^Python 2.7.`

jdemaeyer: Just realized the current test also matches `3.2.0` :see_no_evil:
I'm fine with checking 2.7. I thought Ubuntu Precise was still on 2.6 but even they have upgraded. Why the `^Python`?

dangra: it was a way to say `py_ver.startswith('Python 2.7.')`

jdemaeyer: I'm a little afraid of unnecessarily excluding non-CPython implementations. E.g. what if PyPy outputs "PyPy 2.7.11". But then again I guess it doesn't, it is the language of the version and not the interpreter after all.

PR #144, opened 2016-01-20
Remove overwritten shared library path in patch_sys_executable

jdemaeyer:


shub/utils.py
```diff
@@ -105,9 +105,16 @@ def patch_sys_executable():
...
+        # PyInstaller sets this environment variable in its bootloader. Remove
+        # it so the system-wide Python installation uses its own library path
+        # (this is particularly important if the system Python version differs
+        # from the Python version that the binary was compiled with)
+        orig_lib_path = os.environ.pop('LD_LIBRARY_PATH', None)
         sys.executable = py_exe
         yield
         sys.executable = orig_exe
+        if orig_lib_path:
+            os.environ[str('LD_LIBRARY_PATH')] = orig_lib_path
```
dangra: why `str()`?

jdemaeyer: We imported `unicode_literals` from future and unicode is not allowed in environment variables

dangra: isn't `u''` syntax allowed?

dangra: str() output in py3 will be unicode string, but bytes in py2. :confused: 

and I just tried with this simple script and it works in py2 and py3

```
from __future__ import unicode_literals
import os
os.environ['foo'] = 'bar'
print(os.environ.get('foo'))
```

dangra: I mean, does it work for `os.environ.pop('LD_LIBRARY_PATH', None)` but not for this line?

jdemaeyer: Reading from it is fine b/c of how Python handles string keys:
```
In [1]: a = {}
In [2]: a[str('key')] = 'value'
In [3]: a[unicode('key')]
Out[3]: 'value'
```

I originally had [this script](https://github.com/scrapinghub/shub/blob/9989ee0954a75b687b30160daf22fd5c2d1f788f/freeze/hooks/runtime-hooks.py) in `utils.py` (thus writing unicode to the environment vars) and I remember shub yelling at me when I tried to run a command. Might have been on Windows only

PR #148, opened 2016-01-21
Remove overwritten shared library path in patch_sys_executable

dangra:
~~dummy PR to test a variation of #144 on all platforms~~

It adds a fake api server and integration tests for frozen binary 

Closes #143, closes #144


jdemaeyer:
Hm, interesting. We can leave it out if you want but I wash my hands of responsibility ;)

jdemaeyer:
Ah, wait. The tests never actually run with the frozen binaries. (I.e., the line in question never gets hit, let alone a subprocess being spawned that tries to use the env var)

dangra:
My question stands, why does it work for `os.environ.pop('NAME', None)` (no str()) and it doesn't for `os.environ['NAME']` ?

jdemaeyer:
The first one doesn't write to the environment (or rather, doesn't add or modify).

The problem is not that `os.environ[u'NAME'] = 'xyz'` throws an exception (which it doesn't), but that it leaves the environment in a crippled state, at least on Windows.

Here are two threads on SO that ran into this:
- https://stackoverflow.com/questions/12253014/why-does-popen-fail-on-windows-if-the-env-parameter-contains-a-unicode-object
- https://stackoverflow.com/questions/33551844/environment-can-only-contain-strings-error-while-creating-a-django-cms-project

dangra:
Good, that serves as reference. 
Does it work if we use bytes literals then:: `b'NAME`` ? :)

jdemaeyer:
hehe yeah that looks nicer, I'll update

shub/utils.py
```diff
@@ -94,20 +94,30 @@ def patch_sys_executable():
...
             except NotFoundException:
-                # Explicitly ask for installing 2.7
-                raise NotFoundException("Please install Python 2.7")
+                pass
+        else:
+            raise NotFoundException('Please install Python 2.7')
+
+        output = subprocess.check_output([py_exe, '--version'],
+                                         stderr=subprocess.STDOUT)
+        if not output.startswith('Python 2.7.'):
```
jdemaeyer: Still unsure about `.startswith()`. What if there's warnings being thrown?

jdemaeyer:
Could you thin the testproject to not clutter up the repo? Otherwise LGTM :+1: (except startswith)

dangra:
It contains the default files generated by `scrapy startproject testproject` plus a simple spider so `scrapy list` has something to show in case we need it.

jdemaeyer:
Right right, I just meant all the comments ;)

dangra:
@jdemaeyer check now

shub/utils.py
dangra: Ok, I think you have a point.

dangra: PR updated.

jdemaeyer:
Awesome!

PR #153, opened 2016-01-22
refactored "shub.utils.latest_github_release"

carlosp420:
This change reduced the complexity of "latest_github_release". I also
added a test for the new function "read_release_data".

Also replaced the generic "Exception" to "ValueError".

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -450,11 +450,7 @@ def latest_github_release(force_update=False, timeout=1., cache=None):
                                   'last_release.txt')
     today = datetime.date.today().toordinal()
     if not force_update and os.path.isfile(cache):
-        with open(cache, 'r') as f:
-            try:
-                release_data = json.load(f)
-            except Exception:
-                release_data = {}
+        release_data = read_release_data(cache)
         # Check for equality (and not smaller or equal) so we don't get thrown
         # off track if the clock was ever misconfigured and a future date was
         # saved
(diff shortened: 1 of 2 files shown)


shub/utils.py
```diff
@@ -476,6 +472,16 @@ def latest_github_release(force_update=False, timeout=1., cache=None):
     return release_data
 
 
+def read_release_data(cache):
```
jdemaeyer: Hm I don't really see an advantage in outsourcing this function but I also have no strong opinion on it. ;) If you do outsource it should be generic though, e.g. `def read_json_dict(filename):`

shub/utils.py
```diff
@@ -476,6 +472,16 @@ def latest_github_release(force_update=False, timeout=1., cache=None):
...
+    """Read release data from file. Returns empty dict if data is missing."""
+    with open(cache, 'r') as f:
+        try:
+            release_data = json.load(f)
+        except ValueError:
+            release_data = {}
+    return release_data
+
```
jdemaeyer: I see that I screwed up the error handling in the first place. What I meant to do (and why I used `catch Exception`) is "don't fail when the file can for some reason not be read", i.e. "never fail":
```python
    try:
        with open(cache, 'r') as f:
            release_data = json.load(f)
    except Exception:
        release_data = {}
    return release_data
```

tests/test_utils.py
```diff
@@ -275,6 +275,17 @@ def test_latest_github_release(self, mock_get):
             with open('./cache.txt', 'r') as f:
                 self.assertEqual(f.read(), 'abc')
 
+    def test_read_release_data(self):
+        with open('./cache.txt', 'w') as f:
```
jdemaeyer: This file will not be cleaned from the filesystem after the test. You can use `with self.runner.isolated_filesystem()`

jdemaeyer:
Tentatively closing this, feel free to reopen :)

PR #161, opened 2016-02-08
add command copy-eggs

pawelmhm:
command allows you to download all eggs from old project and upload same eggs
to new project. 

This was requested by @horva 

let me know if current implementation is ok, if it's ok I'll add some tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ dist
 .eggs/
 .tox
 .cache
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 .eggs/
 .tox
 .cache
+.idea
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,2 @@
+mock
+pytest
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,2 +1,0 @@
-mock
-pytest
--- a/shub/copy_eggs.py
+++ b/shub/copy_eggs.py
@@ -1,8 +1,10 @@
 from __future__ import absolute_import
 import os
 from six.moves.urllib.parse import urljoin
+from tempfile import mkdtemp
 import click
 import requests
+from shutil import rmtree
 
 from shub.config import get_target
 from shub.fetch_eggs import fetch_eggs
@@ -19,22 +21,25 @@ project in dash.
 
 
 @click.command(help=HELP, short_help=SHORT_HELP)
-@click.argument("target", required=False, default='default')
-@click.argument("new_project", required=True)
+@click.option("--source_project", prompt="From which projects should I download eggs?")
+@click.option("--new_project", prompt="To which project should I upload eggs?")
 @click.option("-m", "--copy-main", default=False, is_flag=True, help="copy main Scrapy project egg")
-def cli(target, new_project, copy_main):
-    project, endpoint, apikey = get_target(target)
+def cli(source_project, new_project, copy_main):
+    project, endpoint, apikey = get_target(source_project)
+    new_project, new_endpoint, new_apikey = get_target(new_project)
+    copy_eggs(project, endpoint, apikey, new_project, new_endpoint, new_apikey, copy_main)
 
-    copy_eggs(project, endpoint, apikey, int(new_project), copy_main)
 
-
-def copy_eggs(project, endpoint, apikey, new_project, copy_main):
+def copy_eggs(project, endpoint, apikey, new_project, new_endpoint, new_apikey, copy_main):
 
     egg_versions = get_eggs_versions(project, endpoint, apikey)
-    destfile = 'eggs-%s.zip' % project
+    temp_dir = mkdtemp()
+    destfile = os.path.join(temp_dir, 'eggs-%s.zip' % project)
     fetch_eggs(project, endpoint, apikey, destfile)
-    decompress_egg_files()
-    destdir = "eggs-{}".format(project)
+
+    # this will decompress egg containing other eggs, not eggs from source project
+    decompress_egg_files(directory=temp_dir)
+    destdir = os.path.join(temp_dir, "eggs-{}".format(project))
     for egg_name in os.listdir(destdir):
         if egg_name == "__main__.egg" and not copy_main:
             continue
@@ -42,15 +47,17 @@ def copy_eggs(project, endpoint, apikey, new_project, copy_main):
         version = egg_versions[name]
         egg_path = os.path.join(destdir, egg_name)
         egg_info = (egg_name, egg_path)
-        _deploy_dependency_egg(new_project, endpoint, apikey, name=name, version=version,
+        _deploy_dependency_egg(new_project, new_endpoint, new_apikey, name=name, version=version,
                                egg_info=egg_info)
 
+    # remove temporary directory
+    rmtree(temp_dir)
 
 
 def get_eggs_versions(project, endpoint, apikey):
     click.echo('Getting eggs list from project {}...'.format(project))
     list_endpoint = urljoin(endpoint, "eggs/list.json")
-    response = requests.get(urljoin(list_endpoint, "?project={}".format(project)),
+    response = requests.get(list_endpoint, params={"project": project},
                             auth=(apikey, ''))
     response.raise_for_status()
     obj = response.json()
--- a/shub/copy_eggs.py
+++ b/shub/copy_eggs.py
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
 import os
 from six.moves.urllib.parse import urljoin
 import click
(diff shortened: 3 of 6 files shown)


chekunkov:
^ should be solved by 
```
from __future__ import absolute_import
```
in shub/sync_project.py

pablohoffman:
does it also clone the main project egg?

pawelmhm:
@pablohoffman you mean egg called  __main__.egg? no it doest clone it, should it clone it?

@chekunkov good point will fix that

pablohoffman:
If doesn't copy the main egg, `sync-project` is a wrong name to me.

It should either be `sync-eggs` or `sync-deps`, but perhaps we should a add (command line) flag to enable copying the main project egg.

Now, just by looking at the name of other commands (`fetch-eggs`, `deploy-egg`) a more consistent name seems to be `copy-eggs`.

So, in short, rename to `copy-eggs` & add flag for copying main project egg (disabled by default?).

pawelmhm:
yeah copy-eggs sounds good, and optional flag for copying `__main__` is good idea, I would make it disabled by default because perhaps your previous project doesn't have most recent version of your spiders.

will make those changes and update PR

shub/utils.py
```diff
@@ -258,16 +258,20 @@ def build_and_deploy_egg(project, endpoint, apikey):
...
+
+    if egg_info is None:
+        egg_info = _get_egg_info(name)
+    egg_name, egg_path = egg_info
     url = urljoin(endpoint, 'eggs/add.json')
     data = {'project': project, 'name': name, 'version': version}
     auth = (apikey, '')
 
-    click.echo('Deploying dependency to Scrapy Cloud project "%s"' % project)
+    click.echo('Deploying dependency {} {} to Scrapy Cloud project {}'.format(name, version, project))
```
chekunkov: in Python 2 `.format` will raise an unicode encode error if any of input values contain unicode characters and string itself is not unicode. so either use `u'{}'.format(x)` or switch back to `'%s' % x`

pawelmhm: but project should always be integer, I'm forcing it to integer on line 27, so there is no risk here.

chekunkov: right, missed that

pawelmhm:
i spent some time trying to add some tests, but given that it makes heavy use of all other things there would be around 3-4 mocks and patches (I'd have to mock: fetch_eggs, fetch_egg_versions, _deploy_dependency_egg). Its bit difficult to unit test since script integrates other commands and relies on dash responses and behaviors

.gitignore
```diff
@@ -5,3 +5,4 @@ dist
 .eggs/
 .tox
 .cache
+.idea
```
jdemaeyer: Should probably be in your global `.gitignore`? ^^

requirements-dev.txt
```diff
@@ -0,0 +1,2 @@
+mock
```
jdemaeyer: Test requirements are in `tox.ini` and automatically installed when invoking tox. Is there a strong reason to maintain a separate file?

shub/copy_eggs.py
```diff
@@ -0,0 +1,57 @@
...
+
+Fetch all eggs from one project and upload them to other project. This allows
+you to easily clone requirements from old project if you set up some new
+project in dash.
+"""
+
+
+@click.command(help=HELP, short_help=SHORT_HELP)
+@click.argument("target", required=False, default='default')
+@click.argument("new_project", required=True)
```
jdemaeyer: I think putting a required argument after a non-required argument doesn't work as we want. It'll interpret `shub copy-eggs 123` as `target=123` and `new_project=not defined`.

shub/copy_eggs.py
```diff
@@ -0,0 +1,57 @@
...
+@click.option("-m", "--copy-main", default=False, is_flag=True, help="copy main Scrapy project egg")
+def cli(target, new_project, copy_main):
+    project, endpoint, apikey = get_target(target)
+
+    copy_eggs(project, endpoint, apikey, int(new_project), copy_main)
```
jdemaeyer: Users should be able to use named handles (as defined in `scrapinghub.yml`) for the target too, maybe even allowing different endpoints for source and target.
Something like `new_project, _, _ = get_target(new_project)` or `new_project, new_endpoint, new_apikey = get_target(new_project)`

shub/copy_eggs.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def copy_eggs(project, endpoint, apikey, new_project, copy_main):
+
+    egg_versions = get_eggs_versions(project, endpoint, apikey)
+    destfile = 'eggs-%s.zip' % project
+    fetch_eggs(project, endpoint, apikey, destfile)
+    decompress_egg_files()
```
jdemaeyer: Took me a while that this doesn't really decompress eggs, but just decompresses the `eggs-12345.zip` file (containing further eggs), maybe you could add a comment?

More importantly, it would be good if the command cleaned up after itself, e.g. by downloading/decompressing into a temporary directory.

shub/utils.py
```diff
@@ -258,16 +258,20 @@ def build_and_deploy_egg(project, endpoint, apikey):
...
-    egg_name, egg_path = _get_egg_info(name)
-
+def _deploy_dependency_egg(project, endpoint, apikey, name=None, version=None, egg_info=None):
+    if name is None:
+        name = _get_dependency_name()
+    if version is None:
+        version = pwd_version()
+
+    if egg_info is None:
+        egg_info = _get_egg_info(name)
```
jdemaeyer: nitpick: `name = name or _get_dependency_name()` and similar are a little shorter (and implement the same behaviour since we don't allow empty strings as name or version)

shub/copy_eggs.py
```diff
@@ -0,0 +1,57 @@
...
+        egg_info = (egg_name, egg_path)
+        _deploy_dependency_egg(new_project, endpoint, apikey, name=name, version=version,
+                               egg_info=egg_info)
+
+
+
+def get_eggs_versions(project, endpoint, apikey):
+    click.echo('Getting eggs list from project {}...'.format(project))
+    list_endpoint = urljoin(endpoint, "eggs/list.json")
+    response = requests.get(urljoin(list_endpoint, "?project={}".format(project)),
```
jdemaeyer: nitpick: `requests.get(list_endpoint, params = {'project': project}`, ...)` is a little easier to read

shub/copy_eggs.py
```diff
@@ -0,0 +1,57 @@
...
+    copy_eggs(project, endpoint, apikey, int(new_project), copy_main)
+
+
+def copy_eggs(project, endpoint, apikey, new_project, copy_main):
+
+    egg_versions = get_eggs_versions(project, endpoint, apikey)
+    destfile = 'eggs-%s.zip' % project
+    fetch_eggs(project, endpoint, apikey, destfile)
+    decompress_egg_files()
+    destdir = "eggs-{}".format(project)
```
jdemaeyer: Maybe we can make `decompress_egg_files()` return the folder it extracted to, instead of replicating how to path is constructed.

jdemaeyer:
I've added a few comments but looks good overall, thanks! Could you lint the files (in particular, limit lines to a length of 79 characters and unify the usage of empty lines).

Even though I agree doing the three mocks is annoying, I would really prefer to have a test (I know you're busy with other stuff, so I'd be fine with adding it myself). An alternative to mocking would be to change the implementation to the clean architecture, i.e. have a function `deploy_fetched_eggs(eggs_zip, egg_versions)` that returns a list of `(name, version, egg_info)` tuples, and then unit-test that without requiring any mocks.

shub/copy_eggs.py
chekunkov: or accept optional argument to tell where files should be decompressed - this way you can create tmp dir `tempfile.mkdtemp`, pass its path to the function, process eggs, cleanup.

pawelmhm:
thanks for thorough review @jdemaeyer and @chekunkov . I'm going to check comments and update PR bit later this week

shub/copy_eggs.py
pawelmhm: what do you think about not having default project for this command? I think being explicit here is important. So command syntax would be for example:

```
shub copy-eggs 1886 16206
```
with BOTH project ids being required.

Additionaly I think we could add some prompt for user to confirm she wants to deploy eggs FROM one project with id 1886 to other project with id 16206. Will that be ok?

pawelmhm: Another option that I tested now is prompting user for source project and target project, so something like this:

```
> shub copy-eggs                                                                                                                                                    
From which projects should I download eggs?: 1886
To which project should I upload eggs?: 16206
Getting eggs list from project 1886...
```

pawelmhm:
@jdemaeyer @chekunkov I included all your comments, check updated PR

shub/copy_eggs.py
jdemaeyer: I am fine with not having a default project for this command, its usage might indeed become confusing otherwise.

I don't have a strong opinion on the UI. I am generally not a big fan of prompts because they sometimes make automation harder, but this command will probably be called mostly manually (and it can still be automated anyways, yay for `click`!), so lgtm right now :)

jdemaeyer:
Hey @pawelmhm 
Sorry, I somehow missed that you had updated the PR. Looks good to me now. :)

/cc @dangra @chekunkov: With the introduction of kumo, are we discouraging the use of eggs in favour of requirements files? Should we still merge this (maybe undocumented)?

chekunkov:
eggs will exist for some time and we don't know when exactly they will be deprecated. some project will not switch to requirements files instantly and may find this feature useful. so +1 to merge and provide docs - deprecate when we start deprecation cycle for eggs.

PR #168, opened 2016-02-29
Don't migrate non-integer project names

### LINKED ISSUE #149: Don't import non-integer project names
`scrapy startproject myprojectname` puts the following section into `scrapy.cfg`:
```ini
[deploy]
project = myprojectname
```

When users first run `shub deploy`, this will be transferred to `scrapinghub.yml` and an error message will be printed.

We should not import non-integer project names from `scrapy.cfg` and instead guide users through the deploy wizard.

jdemaeyer:
Fixes #149 

Also, `shub deploy` will now also call the deploy wizard when there is a `scrapinghub.yml` without default target.

### DIFF
--- a/shub/config.py
+++ b/shub/config.py
@@ -221,7 +221,7 @@ def _migrate_to_global_scrapinghub_yml():
 
 
 PROJECT_MIGRATION_OK_BANNER = """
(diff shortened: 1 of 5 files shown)


pablohoffman:
functionality LGTM based on the pull request description

PR #169, opened 2016-02-29
Fix PyPI releasing

jdemaeyer:
Currently, there is [an error while Travis tries to deploy shub to PyPI](https://travis-ci.org/scrapinghub/shub/jobs/108800444#L457). The reason is that `python setup.py sdist` and `PyInstaller` both use the `dist` folder, and whatever script deploys to PyPI doesn't like the `shub` executable in `dist`.

That's why there's no source tarball on PyPI for our newer releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
 script: tox -e py27,freeze
 
 before_deploy:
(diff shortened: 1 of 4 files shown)


pablohoffman:
good catch!

PR #172, opened 2016-03-02
Add config option for kumo stacks and requirements file

dangra:


dangra:
@pablohoffman @jdemaeyer I think it is ready to review and get feedback.

shub/config.py
```diff
@@ -164,15 +171,47 @@ def get_version(self):
...
+            project_id=self.get_project_id(target),
+            endpoint=self.get_endpoint(target),
+            apikey=self.get_apikey(target, required=auth_required),
+            stack=self.get_stack(target),
+            requirements_file=self.requirements_file,
         )
 
 
+class Target(object):
+
```
jdemaeyer: Why not a NamedTuple?

shub/config.py
```diff
@@ -164,15 +171,47 @@ def get_version(self):
...
+                 requirements_file):
+        self.name = name
+        self.project_id = project_id
+        self.endpoint = endpoint
+        self.apikey = apikey
+        self.stack = stack
+        self.requirements_file = requirements_file
+
+    def __getitem__(self, pos):
+        return (self.project_id, self.endpoint, self.apikey)[pos]
```
jdemaeyer: This looks like a backwards-compatibility layer for my older screwed-up implementation (should have used `namedtuple` instead of `tuple` in the first place). I would rather scratch this and seize the chance to clean up the codebase, using `target.project_id` instead of `target[0]` everywhere.

shub/config.py
```diff
@@ -164,15 +171,47 @@ def get_version(self):
...
+
+    def __eq__(self, other):
```
jdemaeyer: Can probably also be dropped with `namedtuple`

shub/config.py
```diff
@@ -164,15 +171,47 @@ def get_version(self):
...
+        if isinstance(other, tuple):
+            return tuple(self) == other
+        return self.name == other.name and \
+            self.project_id == other.project_id and \
+            self.endpoint == other.endpoint and \
+            self.apikey == other.apikey and \
+            self.stack == other.stack
```
jdemaeyer: What about `requirements_file`?

shub/config.py
```diff
@@ -164,15 +171,47 @@ def get_version(self):
...
+
+
+
+
```
jdemaeyer: So many blank lines!! :P

shub/config.py
```diff
@@ -28,6 +28,8 @@ def __init__(self):
         }
         self.apikeys = {}
         self.version = 'AUTO'
+        self.stacks = {}
+        self.requirements_file = None
```
jdemaeyer: This was probably discussed elsewhere before: Why not default to `requirements.txt`? I'd think that's what most users expect, and it would facilitate a care-free `shub deploy`.

jdemaeyer:
Thanks @dangra. I would prefer to switch Target containers to `namedtuple` if possible (and scratch using them as tuple everywhere), otherwise lgtm

jdemaeyer:
Oops missed one thing: We've defined `requirements_file` but we never use it in a request to Cloud? Should we also update our `setup.py` template or something?

shub/config.py
dangra: why a tuple at all?

I tried to keep it working without modifying current code too much but I think returning a tuple is just not good solution

shub/config.py
dangra: Me too. But I am aware of some tools expecting `get_target()` to be a tuple like https://github.com/scrapinghub/kumo-release-tool/search?utf8=%E2%9C%93&q=get_target&type=Code

shub/config.py
dangra: I don't think so, the tuple will be longer that comparing only the first 3 elements.

dangra: btw, this is only here because tests cases uses them, we can drop it if we port tests.

shub/config.py
jdemaeyer: I agree that doing a tuple was a bad idea. What I'm saying is that the class you proposed here (which I think is a good idea) can be generated automatically for us by using `namedtuple`.

shub/config.py
dangra: good catch!

shub/config.py
dangra: sorry about that :)

shub/config.py
dangra: Yes, it will cause panic on existent projects that suddenly after upgrading shub will start installing more dependencies than used to be.

I think we can default it to another more uncommon file, like `requirements-scrapinghub.txt`, unlikely it exists and you can pin dependencies in more detail while still can use requirements.txt for local development.

dangra:
> Oops missed one thing: We've defined requirements_file but we never use it in a request to Cloud? Should we also update our setup.py template or something?

Good catch again! I updated the PR to include requirements posting.

shub/config.py
dangra: I see your point, but if we keep the tuple support due to other tools relying on it, the correct process will be to deprecate the tuple-like access and then it requires extending `__getitem__` and `__eq__` anyway.

Another option is to use a different method for "get_target()", like `get_target_conf()`, it returns a flat configuration for target.

jdemaeyer: Damn, we (well, okay, I) just introduced this API and it already falls apart.

Yeah, introducing `get_target_conf()` or similar sounds like the cleanest solution. I can't resist mentioning that especially now that we wouldn't have to extend `__getitem__` or `__eq__`, I would definitely prefer `namedtuple`, if only for the free `__repr__` and keyword args for `__init__` we get with it. However, it's a small detail, this is the last time I mention it and if you insist on writing the class code out instead of using the factory I will not fight you. :P

shub/config.py
jdemaeyer: Let's port tests (I can do it if you're busy, should serve as a good meditation :D)

shub/config.py
chekunkov: let's agree to default to None. if users would like to use `requirements.txt` - that's 1 line in config. `requirements-scrapinghub.txt` is ugly.

shub/config.py
immerrr: https://pypi.python.org/pypi/attrs is a much better way to get free repr & inits, albeit at a price of extra dependency

PR #175, opened 2016-03-21
Add support for stack and reqfile configuration

### LINKED ISSUE #171: Upload requirements.txt for Kumo projects
#We would like to make `shub` upload `requirements.txt` file alongside with the project egg. This will only work on Kumo (Scrapy Cloud 2.0) projects, but will be a much more convenient way to handle project dependencies, than dealing with eggs.

There is a new `requirements` parameter that `addversion` API accepts (in kumo projects) that `shub` should use to upload the requirements file.

To enable this functionality, we also need to add support in `scrapinghub.yml` to declare this requirements file. It should be possible to override per project and the usual override mechanism provided in the YAML conf.

Let's add this feature undocumented for now, to avoid confusion (since it won't work on SC1.0 projects and it may not fail either). We will add documentation once Kumo rolls out to the public in a month from now.

jdemaeyer:
Add support for specifying stack and requirements file per-project using the following syntax:

```
# Aliases for stacks
stacks:
  ocr: ocr:v1.2

projects:
  default: 
    id: 123
    # Specifying stack directly
    stack: scrapy:v1.1
  prod: 1  # Will use Dash's default stack
  ads:
    id: 2
    # Resolves to 'ocr:v1.2'
    stack: ocr
```

Also makes sure `shub deploy` will always deploy the same configuration as `shub deploy 123`.

Resolves #171. Closes #172.

### DIFF
--- a/shub/config.py
+++ b/shub/config.py
@@ -1,6 +1,8 @@
 import contextlib
 import netrc
 import os
+import warnings
+from collections import namedtuple
 
 import click
 import six
@@ -28,6 +30,8 @@ class ShubConfig(object):
         }
         self.apikeys = {}
         self.version = 'AUTO'
+        self.stacks = {}
+        self.requirements_file = None
 
     def load(self, stream):
         """Load Scrapinghub configuration from stream."""
@@ -35,9 +39,11 @@ class ShubConfig(object):
             yaml_cfg = yaml.safe_load(stream)
             if not yaml_cfg:
                 return
-            for option in ('projects', 'endpoints', 'apikeys'):
+            for option in ('projects', 'endpoints', 'apikeys', 'stacks'):
                 getattr(self, option).update(yaml_cfg.get(option, {}))
             self.version = yaml_cfg.get('version', self.version)
+            if 'requirements_file' in yaml_cfg:
+                self.requirements_file = yaml_cfg['requirements_file']
         except (yaml.YAMLError, AttributeError):
             # AttributeError: stream is valid YAML but not dictionary-like
             raise ConfigParseException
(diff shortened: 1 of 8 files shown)


shub/config.py
```diff
@@ -35,9 +39,11 @@ def load(self, stream):
             yaml_cfg = yaml.safe_load(stream)
             if not yaml_cfg:
                 return
-            for option in ('projects', 'endpoints', 'apikeys'):
+            for option in ('projects', 'endpoints', 'apikeys', 'stacks'):
                 getattr(self, option).update(yaml_cfg.get(option, {}))
             self.version = yaml_cfg.get('version', self.version)
+            if 'requirements_file' in yaml_cfg:
```
chekunkov: maybe shorten it to 'requirements'?

shub/config.py
```diff
@@ -35,9 +39,11 @@ def load(self, stream):
...
+                self.requirements_file = yaml_cfg['requirements_file']
```
chekunkov: and it seems like you can have one line here `self.requirements_file = yaml_cfg.get('requirements_file')`

shub/config.py
chekunkov: @jdemaeyer @dangra ^^^ ?

jdemaeyer: I would rather be explicit and stick to `requirements_file`. `requirements` somewhat implies that you could just list requirements, e.g.
```
projects:
  prod:
    id: 123
    requirements:
      - scrapy-fake-useragent
      - boto
```

chekunkov: ok, I won't insist

dangra:
is it possible to fix travis-ci before merging?

shub/config.py
dangra: I coincide with @jdemaeyer that requirements_file is more explicit about the fact it is a pointer to a file location

jdemaeyer:
Seem the Travis errors are not related to this PR, the master branch build fails as well: https://travis-ci.org/scrapinghub/shub/builds/116390454 I'll look into it

PR #180, opened 2016-04-21
Allow setting default stack

jdemaeyer:


pawelmhm:
is this actual failure on windows or just appveyor creating some artificial problems? seems like build error is 

```
WindowsError: [Error 5] Access is denied: 'c:\\users\\appveyor\\appdata\\local\\temp\\1\\pip-omdcen-uninstall\\projects\\shub\\.tox\\freeze\\scripts\\pip.exe'
You are using pip version 8.0.0, however version 8.1.1 is available.
You should consider upgrading via the 'python -m pip install --upgrade pip' command.
```

jdemaeyer:
It's an issue on Windows machines when compiling the frozen binary version of `shub` (so not a real usage issue). You cannot use `pip install -U pip` to upgrade pip, because you cannot overwrite `pip.exe` while it's in use... that has been annoying for the Windows builds for some time now. We got around it by not forcing any specific pip version so far but now that there's a regression in pip 8.1 that's no longer an option. 92cf119 should do it for now.

PR #186, opened 2016-05-17
removing Kumo references

nestortoledo:


jdemaeyer:
Thanks @nestortoledo !

PR #194, opened 2016-07-07
Add flake8 to the Travis build

eliasdorneles:
The motivation is to avoid bikeshedding on the PRs about PEP-8 stuff. :)

I put a relaxed line limit conforming to the current code, seemed good enough.

Does this look good?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,9 @@ deps =
     pytest
     pytest-cov
     flake8
-commands = flake8
+
+commands =
+    flake8
     py.test --cov=shub --cov-report= {posargs:shub tests}
 
 [testenv:freeze]
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,8 @@ deps =
     mock
     pytest
     pytest-cov
-commands =
+    flake8
+commands = flake8
     py.test --cov=shub --cov-report= {posargs:shub tests}
 
 [testenv:freeze]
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -10,7 +10,8 @@ deps =
     mock
     pytest
     pytest-cov
```
jdemaeyer: should `flake8` be listed as dependency here?

eliasdorneles: it is. :)
the diff is confusing.

tox.ini
```diff
@@ -10,7 +10,8 @@ deps =
...
-commands =
+    flake8
+commands = flake8
```
jdemaeyer: Speaking of bikeshredding, could you add a newline before `flake8` here? ;) Almost read over it if it wasn't for the diff highlighting

eliasdorneles: I tried that, but I'm not sure why, the build doesn't fail when `flake8` fails if I move it to the next line.

eliasdorneles: Hm, I can't reproduce it anymore so it was probably supertitious of me. Lemme do the change. :)

tox.ini
eliasdorneles: added a newline so the diff isn't confusing anymore :)

jdemaeyer:
Thanks Elias :) could you rebase?

eliasdorneles:
The build `freeze` seems to have gotten stuck.
In any case, this seems to be pretty safe (passed all other builds), right @jdemaeyer ?

jdemaeyer:
yup, thanks!

PR #195, opened 2016-07-07
tail log output in order to avoid to download the entire log

kalessin:
Note this PR also changes the default behavior (by default will load last 50 lines, not entire log)

Considering the common usage of shub log, I am not sure if we should conserve the default behavior or not.

### DIFF
--- a/shub/log.py
+++ b/shub/log.py
@@ -38,9 +38,11 @@ SHORT_HELP = "Fetch log from Scrapy Cloud"
 
 class LogIterFunc(object):
     def __init__(self, job, tail):
-        logstats = job.logs.stats()
-        lines = logstats['totals']['input_values']
-        lastline = max(lines - tail, -1)
+        lastline = -1
+        if tail is not None:
+            logstats = job.logs.stats()
+            lines = logstats['totals']['input_values']
+            lastline = max(lines - tail, -1)
         if lastline == -1:
             lastline = None
         self._startafter = None if lastline is None else '{}/{}'.format(job.key, lastline)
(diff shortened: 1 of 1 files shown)


eliasdorneles:
When using just `shub log JOB_ID` I'd expect the default behavior of downloading the whole log.

Probably best to limit only when `--tail`.

kalessin:
Just changed default behavior. Now it downloads everything if you don't use -n

eliasdorneles:
Nice, looks good to me! Thanks, @kalessin :+1:

shub/log.py
```diff
@@ -36,13 +36,33 @@
 SHORT_HELP = "Fetch log from Scrapy Cloud"
 
 
+class LogIterFunc(object):
```
jdemaeyer: This looks like a big class just for setting an initial `startafter`. Maybe we can move the "Determine first `_startafter`" logic into `cli()` and allow a `startafter` keyword arg to `job_resource_iter`?

kalessin: That was my first intention, but then I found that implied an important refactoring in the core of shub. Also, I don't see the problem with this class. Conserves separation of functions and generality of the core shub functions.

kalessin: @jdemaeyer any concern with that approach?

jdemaeyer: Hey @kalessin Sorry for the huge delay :/

I would really rather avoid this class. I believe an "initial `startafter`" feature is very much generic enough to fit into the `job_resource_iter` core function. We may after all want to use it for `requests` or `items` too at some point (dunno if that's supported by the JobQ API).

PR #196, opened 2016-07-11
Add py3 support

### LINKED ISSUE #188: Add py3 support
About time ;)

immerrr:
Should close #188.


appveyor.yml
```diff
@@ -2,34 +2,41 @@ platform: x86
...
 after_test:
-  - 7z a dist_bin\shub-%APPVEYOR_REPO_TAG_NAME%-windows-%PLATFORM%.zip %APPVEYOR_BUILD_FOLDER%\dist_bin\shub.exe
+  - 7z a dist_bin\shub-%APPVEYOR_REPO_TAG_NAME%-windows-%PLATFORM%-%TOX_ENV%.zip %APPVEYOR_BUILD_FOLDER%\dist_bin\shub.exe
 
 deploy:
   provider: GitHub
   auth_token:
     secure: mNA38grRkwGjw4zSUX1qA67eE6CRPjk97hRngZf19FuopBwbpPjOALBIlYZWKlDv
-  artifact: dist_bin\shub-$(APPVEYOR_REPO_TAG_NAME)-windows-$(PLATFORM).zip
+  artifact: dist_bin\shub-$(APPVEYOR_REPO_TAG_NAME)-windows-$(PLATFORM)-%TOX_ENV%.zip
```
jdemaeyer: Is there a reason to upload two different binaries? They were frozen using different python versions but the functionality should be the same, right?

immerrr: TBH, appveyor and windows artifacts are all new to me, and I don't think I made it work in the end. Feel free to consider this PR a proof-of-concept or a baseline and tweak it to your liking.

shub/utils.py
```diff
@@ -136,12 +137,12 @@ def patch_sys_executable():
         # it so the system-wide Python installation uses its own library path
         # (this is particularly important if the system Python version differs
         # from the Python version that the binary was compiled with)
-        orig_lib_path = os.environ.pop(b'LD_LIBRARY_PATH', None)
+        orig_lib_path = os.environ.pop('LD_LIBRARY_PATH', None)
```
jdemaeyer: Is there a good reason for making this a unicode instead of bytes literal? This will [break the Windows binary](https://github.com/scrapinghub/shub/pull/148#issuecomment-173417050).

shub/utils.py
```diff
@@ -136,12 +137,12 @@ def patch_sys_executable():
...
         sys.executable = py_exe
         yield
         sys.executable = orig_exe
         if orig_lib_path:
-            os.environ[b'LD_LIBRARY_PATH'] = orig_lib_path
+            os.environ['LD_LIBRARY_PATH'] = orig_lib_path
```
jdemaeyer: (as above)

tests/test_jobresource.py
```diff
@@ -16,6 +18,7 @@ def _test_prints_objects(self, cmd_mod, resource_name):
         jobid = '1/2/3'
         with mock.patch.object(cmd_mod, 'get_job', autospec=True) as mock_gj:
             # Patch job.items.iter_json() to return our objects
+            mock_gj.return_value._metadata_updated = time.time()
```
jdemaeyer: Is this related to py3 compatibility?

shub/utils.py
immerrr: It's not unicode, but "native strings". Putting byte-strings breaks py3:
```
In [4]: os.environ[b'foobar']
...
TypeError: str expected, not bytes
```

jdemaeyer:
Hey @immerrr, thanks for the PR! Looks good, I've added a few questions we should settle before merging. :)

Is this still WIP?

shub/utils.py
jdemaeyer: Hm but isn't it unicode because we have `from __future__ import unicode_literals` at the top?

tests/test_jobresource.py
immerrr: There was a problem with mocks that got uncovered when being ported to py3, can't remember now, let me see if i can find it quickly.

shub/utils.py
jdemaeyer: Would it be enough to wrap it in `to_native_str()`?

immerrr: Right, that should not be there,  unicode_literals almost never work properly. I'll fix this.

immerrr: This example is actually a prime example of why unicode_literals work wrong.

tests/test_jobresource.py
immerrr: It worked before because Py2 allowed more slack in comparison operators, in Py3 the following failed:
```
File "/home/immerrr/scrapinghub/shub/shub/utils.py", line 454, in job_live 
    if time.time() - job._metadata_updated > refresh_meta_after:
TypeError: unorderable types: MagicMock() > int()
```

shub/utils.py
immerrr: > Would it be enough to wrap it in to_native_str()?

Well, yes, but you only need to do that for variables that may come in binary or unicode. Literals without any prefix should already be of "native" type.

immerrr:
It is WIP in terms of CI integration. I'm not quite well versed with tox, and have no experience with appveyor. May I kindly ask you to take it from here?

immerrr:
Oh, that thing is in my private repo, that might be a problem. Tests do seem to pass on Windows, so the library code is good, but there are problems "freezing" the binary, so another option would be to merge it now and figure out windows builds afterwards.

PR #198, opened 2016-07-27
Added: Warnings when an API endpoint is using staging.scrapinghub.com or HTTP

starrify:
This PR adds warnings in two cases:
1. When an API endpoint uses `staging.scrapinghub.com`. See also https://scrapinghub.atlassian.net/browse/SC-725
2. When an API endpoint uses HTTP. See also https://scrapinghub.slack.com/archives/security/p1469507392000003

Personally I even wanted to directly "correct" those inappropriate endpoints, since we might assume that users of `shub` are all accessing only Scrapinghub's API (..can we?). But I'm afraid that might look too aggressive.

### DIFF
--- a/shub/config.py
+++ b/shub/config.py
@@ -33,6 +33,30 @@ class ShubConfig(object):
         self.stacks = {}
         self.requirements_file = None
 
+    def _check_endpoints(self):
+        """Check the endpoints. Send warnings if necessary."""
+        for endpoint, url in self.endpoints.items():
+            parsed = six.moves.urllib.parse.urlparse(url)
+            if parsed.netloc == 'staging.scrapinghub.com':
+                self.endpoints[endpoint] = six.moves.urllib.parse.urlunparse(
+                    parsed._replace(netloc='app.scrapinghub.com')
+                )
+                click.echo(
+                    'WARNING: Endpoint "%s" is still using %s which has been '
+                    'obsoleted. Shub has had it updated to app.scrapinghub.com '
+                    'for only this time. Please update the configuration '
+                    'ASAP.' % (
+                        endpoint, parsed.netloc,
+                    ),
+                    err=True
+                )
+            if parsed.scheme == 'http':
+                click.echo(
+                    'WARNING: Endpoint "%s" is still using HTTP. '
+                    'Please change it to HTTPS if possible.' % endpoint,
+                    err=True
+                )
+
     def load(self, stream):
         """Load Scrapinghub configuration from stream."""
         try:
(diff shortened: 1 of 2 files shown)


shub/config.py
```diff
@@ -3,6 +3,7 @@
 import os
 import warnings
 from collections import namedtuple
+import urlparse
```
jdemaeyer: Could you import this through `six`?

shub/config.py
```diff
@@ -33,6 +34,25 @@ def __init__(self):
...
+    def _check_endpoints(self):
+        """Check the endpoints. Send warnings if necessary."""
+        for endpoint, url in self.endpoints.items():
+            parsed = urlparse.urlparse(url)
+            if parsed.netloc == 'staging.scrapinghub.com':
+                click.echo(
+                    'WARNING: Endpoint "%s" is still using %s. '
+                    'Please change the netloc to app.scrapinghub.com' % (
+                        endpoint, parsed.netloc,
+                    ),
```
jdemaeyer: Could you add an explanation that `staging.scrapinghub.com` was dropped/deprecated/no longer supported to the message?

jdemaeyer:
Hey @starrify, thanks for the PR, and apologies for the huge delay :<

> Personally I even wanted to directly "correct" those inappropriate endpoints

I think we can safely auto-correct `staging.scrapinghub.com`, but we should only warn when using HTTP, in case some development environments (for our devs working on Cloud etc.) may not support HTTPS.

While we're at it, could you make this change to `shub/utils.py`, line 60:
```python
-            error = rsp.text
+            error = rsp.text or "Status %d" % rsp.status_code
```

starrify:
Hi @jdemaeyer , thanks a lot for the notes! I've got this PR updated :)

shub/config.py
```diff
@@ -33,6 +33,30 @@ def __init__(self):
...
+        for endpoint, url in self.endpoints.items():
+            parsed = six.moves.urllib.parse.urlparse(url)
+            if parsed.netloc == 'staging.scrapinghub.com':
+                self.endpoints[endpoint] = six.moves.urllib.parse.urlunparse(
+                    parsed._replace(netloc='app.scrapinghub.com')
+                )
+                click.echo(
+                    'WARNING: Endpoint "%s" is still using %s which has been '
+                    'obsoleted. Shub has had it updated to app.scrapinghub.com '
+                    'for only this time. Please update the configuraitions '
```
jdemaeyer: `configuration` :)

starrify: Fixed :)

jdemaeyer:
Thanks @starrify! :)

PR #199, opened 2016-08-08
Validate requirements

lsenta:
This is a proposal to validate requirements.

When you setup dependencies for your SC2.0 project, there are a bunch of edge cases to consider.
This PR adds a validation step for the requirements.txt with good practices described in:
https://support.scrapinghub.com/topics/1970-deploying-dependencies-to-your-scrapinghub-project/

When you `shub deploy`, requirements are validated:

```
› shub deploy                                                                                     ☺
Packing version 1.0
validating requirements:
WARNING: requests has no version specification (l.1)
WARNING: js2xml has no version specification (l.2)
WARNING: extruct points to a git repository, use http scheme (l.5)
ERROR: packagex is editable (l.6)
```

In case of errors, the command stops. In case of warnings, an input appears:
```
project has warning, continue anyway? [y/N]
```

TODOs:
- [ ] exit with an error code
- [ ] link to the relevant documentation on warnings/errors (the support page is fine?)
- [ ] add test of the integration in the endpoints (only validation module is tested)
- [ ] check that integration is correct and enough (only deploy?)

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -2,6 +2,8 @@ click
 requests
 ruamel.yaml
 six
+requirements-parser
 
 hubstorage
 scrapinghub
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ ruamel.ordereddict==0.4.9  # via ruamel.yaml
 ruamel.yaml==0.10.15
 scrapinghub==1.7.0
 six==1.10.0
+requirements-parser==0.1.0
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -8,6 +8,9 @@ from subprocess import check_call
 
 import click
 import setuptools  # not used in code but needed in runtime, don't remove!
+
+from shub.validation import validate_and_dump_for_path
+
 _ = setuptools  # NOQA
 from scrapinghub import Connection, APIError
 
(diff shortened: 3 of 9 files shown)


lsenta:
If you guys think this provide value, please give me some guidance on where to integrate nicely with the tests & the regular flow (TODOs).

chekunkov:
I like the idea, though after some thinking I have doubts that client is a good place for it. E.g. if we add support for private repositories and/or editable pip install - users with old version of shub will not be able to use that right away without upgrading shub - which is not always convenient. Or if we add some new validation rule - users with old shub will never know. AFAIK we don't force users to use the latest shub version. maybe we should add validation on the server side right away and simply report validation errors from response? implementation is simple anyway.

thoughts @pablohoffman @jdemaeyer ?

if above concerns are not valid - client side validation sounds good to me.

jdemaeyer:
I agree with @chekunkov here. When uploading an invalid `requirements.txt`, users should be informed of their error regardless of the `shub` version that they have installed. The only way to achieve this is to have the validation server-side.

Furthermore, `shub` is really not the tool that handles requirements. All it does is upload a file without ever looking at its contents. As cloud is responsible for building stacks, it should also be responsible for validating the recipes of these stacks. Especially since requirements towards `requirements.txt` can change with updates to Cloud, and in my eyes it makes little sense to have to maintain two versions of these requirements, one in Cloud and one in `shub`.

jdemaeyer:
Oh, and of course, thanks @lsenta for the PR! :)

PR #201, opened 2016-09-01
Add ability to send 3rd party eggs when deploying

krotkiewicz:


rolando:
Would be possible to add source packages?

I would love to be able to specify wheels and source packages, like:
```
requirements:
    packages:
        - private/package.whl
        - private/dep.tgz
        # wildcards support too!
        - vendor/*.egg
```

Hopefully `pip install <packages>` will deal with all dependencies. However, supporting only binary packages could lead to errors due to platform mismatch (i.e.: compiling a `.whl` in `osx` while SC container runs `linux-64`).

FWIW, `anaconda-client` does this check to ensure a `.tgz` is a pypi package https://github.com/Anaconda-Platform/anaconda-client/blob/eed3229c4b322d8b12b6b99c6565bc8038ff7c30/binstar_client/utils/detect.py#L92

krotkiewicz:
@rolando of course but first we would like to add support for eggs, then we can extend it to any whl and others.

Scrapycloud currently allows you to add eggs via `Code & Deploys` we just want to move this functionality to shub.

shub/deploy.py
```diff
@@ -128,14 +128,15 @@ def _url(endpoint, action):
...
 
 def _upload_egg(endpoint, eggpath, project, version, auth, verbose, keep_log,
-                stack=None, requirements_file=None):
+                stack=None, requirements_file=None, eggs=None):
+    eggs = eggs or []
     data = {'project': project, 'version': version}
     if stack:
         data['stack'] = stack
-    files = {'egg': ('project.egg', open(eggpath, 'rb'))}
+    files = [('egg', open(path, 'rb')) for path in eggs]
```
vshlapakov: As a proposal: maybe catch `IOError` here and raise `ShubException` with some nice message, instead of :point_down: ?

``` 
 File "/home/kris/ws/schub/sh_tools/shub/shub/deploy.py", line 136, in _upload_egg
    files = [('egg', open(path, 'rb')) for path in eggs]
IOError: [Errno 2] No such file or directory: 'boto/dist/boto-2.42.0-py2.7.egg1'```

**Comment by vshlapakov** [comment]
LGTM, tested locally. +1 for wheels/sources, but it's for another PR for sure.

File: tests/test_config.py [hunk 2]
```diff
@@ -319,11 +326,11 @@ def test_get_target_conf_custom_defaults(self):
         """)
         self.assertEqual(
             self.conf.get_target_conf('shproj'),
-            _target(123, apikey='key', stack='custom_default')
+            _target(123, apikey='key', stack='custom_default', eggs=[])
```
vshlapakov: I'd add some additional tests for deploy procedure.

shub/deploy.py
jdemaeyer: :+1:

tests/test_config.py
jdemaeyer: Yeah, we should at least check if it reads/sends the data from `egg1.egg` and `egg2.egg` properly in `test_deploy`. That's also missing for the `requirements_file` parameter. If you want and have time you can add them to this PR @krotkiewicz, otherwise I can make a follow-up PR.

shub/deploy.py
jdemaeyer: The JIRA ticket says: "When deploying send [the eggs] in `eggs` field. The main egg should be still send in `egg` field". But now apparently we're sending everything in the `egg` field. Were the specs changed? How does Dash know which one is the project egg?

shub/config.py
```diff
@@ -131,6 +133,8 @@ def save(self, path=None):
             yml['apikeys'] = self.apikeys
             yml['version'] = self.version
             yml['stacks'] = self.stacks
+            if self.eggs:
+                yml['requirements'] = {'eggs': self.eggs}
```
jdemaeyer: Hm, I'm not super-fond of having both `requirements` and `requirements_file`. Should we deprecate the `requirements_file` parameter and instead use a `file` string or `files` list under the `requirements` dict? Then again it sucks to already deprecate the quite fresh `requirements_file` parameter. (/cc @dangra)

jdemaeyer:
Thanks Konrad! I agree with the "wheels = awesome but in another PR" sentiment

shub/deploy.py
krotkiewicz: This was a bug here in shub command

tests/test_config.py
krotkiewicz: @jdemaeyer added

shub/config.py
dangra: I was trying to avoid that discussion, but yes the plan should be to deprecate requirement_file, just not sure we want to do that now.

jdemaeyer: Makes sense to do it in another PR. I'll add one in a bit. Or do you think it should also be in a later release?

dangra: If you can add it for this release, that would be awesome.

jdemaeyer:
Thanks @krotkiewicz!

Issue #213, opened 2016-10-22
Add escaping to auto-generated version strings


631068264:
```
Packing version a71f746-feature/pic
Deploying to Scrapy Cloud project "112954"
Error: Deploy failed (400):
version: This value does not match the required pattern.
```

What is **required pattern** ?

jdemaeyer:
Hey @631068264,
thanks for the report!

Version strings need to follow [Docker's limitations on tag names](https://docs.docker.com/engine/reference/commandline/tag/). In your case, it struggles with the forward slash in the branch name. For now, you can do one of the following (most preferable first):
1. [rename the branch](https://gist.github.com/lttlrck/9628955), or
2. specify a custom version string during deploy: `shub deploy --version feature_pic`, or
3. specify a custom version string in your `scrapinghub.yml`: `version: feature_pic` (on the top level)

In the long run, `shub` should properly escape forbidden characters from branch names when auto-generating version strings. At the very least, we should return a more helpful error message.

jdemaeyer:
closed via #215

PR #215, opened 2016-11-11
Clean version string with pattern

vshlapakov:
Dash logic has pattern matching to check correctness of version string on deploy, and in some cases will deploy reject requests from shub tool due to spaces or some other forbidden chars (for example SVN allows using spaces in a branch name). Let's polish clients experience and clean version string before doing any external requests.
 
Review, please.


jdemaeyer:
Thanks @vshlapakov (in particular for adding tests!)

LGTM :) I'll fix the error with Travis OSX freezing, doesn't seem related to this

Closes #213

PR #218, opened 2016-11-22
allow to specify glob patterns in requirements eggs section

kalessin:


jdemaeyer:
Thanks Martin! Ignore the unrelated PEP8 error (Guido decided that [we've been doing it wrong all along](https://github.com/PyCQA/pycodestyle/issues/400) ;)), I'll fix that in `master`.

Can you make it a little less cryptic and add a test please? E.g.
```python
[filename
 for pattern_matches in (glob.glob(e) for e in eggs or [])
 for filename in pattern_matches]
```

PR #219, opened 2016-11-24
Merge shub-image as shub.image

vshlapakov:
The PR merges shub-image library with keeping its history.

Most important things of this PR:
- moving `shub_image` logic under `shub.image` namespace
- extending click logic to call commands with `shub image` prefix
- adding `docker-py`, `retrying` as an additional shub dependency
- copying shub-image docs under docs folder
- merging a lot of auxiliaries files in favour of `shub`

Review, please.


LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (C) 2016 Scrapinghub, Inc
```
vshlapakov: Not sure if we need to add the similar LICENSE for `shub`, don't we?

setup.py
```diff
@@ -21,6 +21,7 @@
     zip_safe=False,
     install_requires=['click', 'hubstorage', 'pip', 'requests', 'PyYAML',
                       'scrapinghub', 'six'],
+    extras_require={'docker-py': ['docker-py>=1.10.0']},
```
vshlapakov: It makes sense to keep using `shub` tool w/o using Docker at all, so `docker-py` should be an optional package here, I'll return a check for docker-py import to disable some click commands.

vshlapakov:
@jdemaeyer I think it's ready for the first initial review, check when you have time please.

shub/tool.py
```diff
@@ -52,3 +53,6 @@ def cli():
     command_module = importlib.import_module(module_path)
     command_name = command.replace('_', '-')  # easier to type
     cli.add_command(command_module.cli, command_name)
+
+# Add image commands group
+cli.add_command(image_cli, "image")
```
jdemaeyer: Wouldn't this work without extra code (i.e. by just adding `image` to the `commands` list) if we rename `image_cli` to `cli` in `shub/image/__init__.py`?)

setup.py
```diff
@@ -36,5 +36,5 @@
         'Topic :: Internet :: WWW/HTTP',
     ],
     test_suite='tests',
-    tests_require=['mock']
+    tests_require=['mock', 'pytest']
```
jdemaeyer: Mhh since we use `tox` to manage testing [and test dependencies](https://github.com/scrapinghub/shub/blob/master/tox.ini#L8), should we just remove this line and the one above? `python setup.py test` fails on my machine.

jdemaeyer: Aren't `python setup.py test` and `tox` two alternative entry points to running our tests, and we should pick one of them?

Related: https://github.com/tox-dev/tox/issues/13

redapple: They're alternatives indeed. We "had" to add that for Debian packaging. shub may not need it, that's true

shub/image/__init__.py
```diff
@@ -0,0 +1,24 @@
+import click
+import importlib
+
+
+@click.group(help="Release project with Docker")
```
vshlapakov: Yeah, naming is the most complex problem as usual 😄 
Maybe `Manage project based on custom Docker image`?

vshlapakov:
@jdemaeyer Thanks for the great review, I think I covered all the issue except a one minor about sound naming for shub image menu item 🙂 Could you check it once again please?

shub/image/__init__.py
jdemaeyer: Yeah, that should be clear enough

jdemaeyer:
Thanks again @vshlapakov! Could you squash the "merge shub-image into shub" commits into two or three commits? Only if that doesn't cost too much time in rebase hell, otherwise just squash them all into one ;)

vshlapakov:
@jdemaeyer Done, let me know if there's anything else to fix / improve  👍

PR #225, opened 2017-01-04
add requirements-test.txt file with test deps

chekunkov:
it's convenient to run `pip install -r requirements.txt -r tests/requirements.txt` locally if you want test deps to be installed together with package deps


jdemaeyer:
I wonder if we should pin dependencies like we do with `requirements.in` -> `requirements.txt`, or if that is disencouraged for test environment packages

chekunkov:
I'm okay to pin versions in requirements-test.txt.

Talking about requirements-test.in - I'm afraid this will only complicate dependency management without gaining any benefits. E.g. one will have to compile both files if any of them changes, otherwise there could be differences between packages in requirements.txt and requirements-test.txt, which will lead to significant differences between test and build environment (note different versions of `requests` and `websocket-client` below).

```diff
→ diff requirements.txt requirements-test.txt
5c5
< #    pip-compile --output-file requirements.txt requirements.in
---
> #    pip-compile --output-file requirements-test.txt requirements-test.in
7,9c7,12
<
< backports.ssl-match-hostname==3.5.0.1  # via docker-py
< click==6.6
---
> appdirs==1.4.0            # via setuptools
> backports.ssl-match-hostname==3.5.0.1  # via docker-py, websocket-client
> backports.ssl-match-hostname==3.5.0.1  # via docker-py, websocket-client
> click==6.7
> configparser==3.5.0       # via flake8
> coverage==4.3.4           # via pytest-cov
12,13c15,28
< hubstorage==0.23.1
< ipaddress==1.0.17         # via docker-py
---
> enum34==1.1.6             # via flake8
> flake8==3.3.0
> funcsigs==1.0.2           # via mock
> ipaddress==1.0.18         # via docker-py
> mccabe==0.6.1             # via flake8
> mock==2.0.0
> packaging==16.8           # via setuptools
> pbr==1.10.0               # via mock
> py==1.4.32                # via pytest
> pycodestyle==2.3.1        # via flake8
> pyflakes==1.5.0           # via flake8
> pyparsing==2.1.10         # via packaging
> pytest-cov==2.4.0
> pytest==3.0.6
15c30
< requests==2.10.0
---
> requests==2.13.0
19c34,38
< websocket-client==0.37.0  # via docker-py
---
> websocket-client==0.40.0  # via docker-py
>
> # The following packages are commented out because they are
> # considered to be unsafe in a requirements file:
> # setuptools                # via pytest
```

where requirement-test.in looks like 

```
-r requirements.in

mock
pytest
pytest-cov
flake8
```

jdemaeyer:
We could imitate the way [scrapy handles this](https://github.com/scrapy/scrapy/blob/master/tox.ini#L10-L16),

1. Move the test requirements to `tests/requirements.txt`
2. Remove the `-r requirements` line in `tests/requirements.txt`
3. Change the `deps` section in `tox.ini` to:
    ```
    deps =
        -rrequirements.txt
        -rtests/requirements.txt
    ```

Then additionally we could pin the test dependency versions via pip-tools and `tests/requirements.in` without running into the issues you raised above.

chekunkov:
```
deps =
    -rrequirements.txt
    -rtests/requirements.txt
```

^ this will lead to 2 separate calls to pip install, if there are any conflicts between dependencies (e.g. if one package in `requirements.txt` and another package in `tests/requirements.txt` happen to have a common dependency, but require different versions) - `pip install tests/requirements.txt` will override dependencies installed by `pip install requirements.txt`.

chekunkov:
why simply pinning versions in requirements-test.txt doesn't work for you and what benefits do you see in using pip-tools for test requirements?

jdemaeyer:
I don't mind not pinning test requirements (or pinning them directly in the .txt file) and not using pip-tools.

But I would like to avoid having `-r requirements.txt` in `tests/requirements.txt`, A) because it obscures what's really going on when you run tox, and B) because we would be in line with the scrapy repo. We can change the order in the `deps =` var if we're concerned about conflicting versions, but I don't think it is a big problem (and one that certainly should be noticed when reviewing PRs).

chekunkov:
sorry, I don't completely follow.

> because it obscures what's really going on when you run tox

why? IMO it's equivalent to saying `pip install -r requirements.txt` obscures what dependencies the project has. a person will look into the tox and will see requirement-test.txt file in deps - so he'll learn that to run tests locally without tox he need to install packages from requirements-test.txt (which is placed near requirements.txt, so there's a good chance that one will make an educated guess without even looking into tox.ini), and he can do that with a single command `pip install -r requirements-test.txt`. that's the main reason why I created this PR, it brings obvious improvement to the development process.

> because we would be in line with the scrapy repo

why is that a reason/goal/benefit when it comes to tests setup and making it easier to install dev environment locally? Scrapy had its own reasons to split test dependencies in such a way, I don't think that this project should follow that approach blindly. unless you have a good reason behind.

> We can change the order in the deps = var if we're concerned about conflicting versions, but I don't think it is a big problem (and one that certainly should be noticed when reviewing PRs).

disagree here, it's easy to miss, unless you *always* keep in mind that this can happen and *always* check that both files don't have conflicting versions for potential repeated value. also if you have only one file changed in the PR - you'll not see if this change brings differences between the two files and you'll have to go to the source code and deliberately compare to the another file which is not in the PR diff.

jdemaeyer:
> why? IMO it's equivalent to saying pip install -r requirements.txt

Because it adds the second step of "look into tox.ini -> look into requirements-tests.txt **-> look into requirements.txt**". Suddenly `tox.ini` no longer contains a reference to what will be installed, but a reference to a reference of what will be installed.

> that's the main reason why I created this PR, it brings obvious improvement to the development process.

`pip install -r tests/requirements.txt -r requirements.txt` is also a single command ;) and from the name alone it is not clear that `requirements-tests.txt` will also install package dependencies. Explicit is better than implicit.

> why is that a reason/goal/benefit when it comes to tests setup and making it easier to install dev environment locally?

Because conventions are important, and more often than not the most important thing is not what the convention is, but that it's actually followed. People that work on this repo have a higher likelihood to also have worked on (or at least checked out) the scrapy repo, or frontera, or scrapyd, or scrapy-inline-requests, or scrapy-redis, all of which do not include `requirements.txt` in their `tests/requirements.txt` (or `requirements-tests.txt`), but list them on two lines in `tox.ini`.

On the third point, see my edit above. pip will complain loudly if there is a version clash.

chekunkov:
> pip install -r tests/requirements.txt -r requirements.txt is also a single command ;)

good point. and actually I was wrong, multiple requirements files in tox are merged and installed in a single command. so I'm okay with removing -r line from test requirements and moving it to `tests` package.

jdemaeyer:
Thanks Alex =)

PR #226, opened 2017-01-04
Add -h option to get help message

chekunkov:
It's a very popular option for getting a help message, can't help but use it, it's already a muscle memory I guess. So annoying that click doesn't have it by default.


jdemaeyer:
Can't believe this isn't default in click. Thanks Alex!

PR #229, opened 2017-01-19
image test: drop check for sh-ep-scrapy

vshlapakov:
Drop redundant check for sh-ep-scrapy.

Initial intention before https://github.com/scrapinghub/shub/pull/229#issuecomment-276886774 was to check for `scrapinghub-entrypoint-scrapy` only if there's Scrapy installed, but it doesn't play nice with non-python code.

I also noticed that we don't clean temporary containers after tests, let's improve it as well.

### DIFF
--- a/shub/image/test.py
+++ b/shub/image/test.py
@@ -70,10 +70,14 @@ def _check_image_exists(image_name, docker_client):
 def _check_sh_entrypoint(image_name, docker_client):
     """Check that the image has scrapinghub-entrypoint-scrapy pkg"""
     status, logs = _run_docker_command(
-        docker_client, image_name,
-        ['pip', 'show', 'scrapinghub-entrypoint-scrapy'])
-    if status != 0 or not logs:
-        raise shub_exceptions.NotFoundException(SH_EP_SCRAPY_WARNING)
+        docker_client, image_name, ['pip', 'show', 'Scrapy'])
+    # doesn't make sense to check sh-ep-scrapy if there's no Scrapy
+    if status == 0 and logs:
+        status, logs = _run_docker_command(
+            docker_client, image_name,
+            ['pip', 'show', 'scrapinghub-entrypoint-scrapy'])
+        if status != 0 or not logs:
+            raise shub_exceptions.NotFoundException(SH_EP_SCRAPY_WARNING)
 
 
 def _check_list_spiders_entry(image_name, docker_client):
@@ -99,10 +103,13 @@ def _check_start_crawl_entry(image_name, docker_client):
 def _run_docker_command(client, image_name, command):
     """A helper to execute an arbitrary cmd with given docker image"""
     container = client.create_container(image=image_name, command=command)
-    client.start(container)
-    statuscode = client.wait(container=container['Id'])
-    logs = client.logs(container=container['Id'], stdout=True,
-                       stderr=True if statuscode else False,
-                       stream=False, timestamps=False)
-    utils.debug_log("{} results:\n{}".format(command, logs))
-    return statuscode, logs
+    try:
+        client.start(container)
+        statuscode = client.wait(container=container['Id'])
+        logs = client.logs(container=container['Id'], stdout=True,
+                           stderr=True if statuscode else False,
+                           stream=False, timestamps=False)
+        utils.debug_log("{} results:\n{}".format(command, logs))
+        return statuscode, logs
+    finally:
+        client.remove_container(container)
--- a/shub/image/test.py
+++ b/shub/image/test.py
@@ -12,22 +12,20 @@ A command to test an image after build step to make sure it fits contract.
 It consists of the following steps:
 
 1) check that image exists on local machine
-2) check that image has scrapinghub-entrypoint-scrapy python package
-3) check that image has start-crawl entrypoint
-4) check that image has list-spiders entrypoint
-
-These entrypoints are provided by scrapinghub-entrypoint-scrapy package,
-so the goal of the last checks is to validate the package version.
+2) check that image has start-crawl entrypoint
+3) check that image has list-spiders entrypoint
 
 If any of the checks fails - the test command fails as a whole. By default,
 the test command is also executed automatically as a part of build command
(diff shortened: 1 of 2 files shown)


shub/image/test.py
```diff
@@ -105,4 +109,5 @@ def _run_docker_command(client, image_name, command):
                        stderr=True if statuscode else False,
                        stream=False, timestamps=False)
     utils.debug_log("{} results:\n{}".format(command, logs))
+    client.remove_container(container)
```
chekunkov: maybe create a container with `--rm`? if things will go wrong above this line container will not be removed

alternatively make it a context manager with try: finally:

vshlapakov: Good idea, but there's no analogue of `docker run` in `docker-py==1.10.6`, it appears only in [`docker-py==2.0`](https://github.com/docker/docker-py/blob/2.0.0/docker/models/containers.py#L422) version - and it requires testing. Considering above I'd go with a context manager for the PR

chekunkov: > create_container
> Creates a container that can then be .start() ed. Parameters are similar to those for the docker run command except it doesn't support the attach options (-a).

@vshlapakov I think you can pass `auto_remove` as a `host_config`

> The arguments that are passed directly to this function are host-independent configuration options. Host-specific configuration is passed with the host_config argument. You’ll normally want to use this method in combination with the create_host_config() method to generate host_config.

vshlapakov: @chekunkov Mm, it doesn't work for me, and looks like [not supported](http://docker-py.readthedocs.io/en/1.10.0/hostconfig/) - author of `docker-py` [recommends](https://github.com/docker/docker-py/issues/1163) using `create-wait-remove` chain instead everywhere.

By the way, TIL `--rm` logic was implemented on client side initially and only now moved to server side in [latest docker](https://github.com/docker/docker/pull/20848).

chekunkov: ok, let's implement as a context manager then

chekunkov: @vshlapakov sorry, just realized that context manager is pointless, we just need try: finally:

vshlapakov: Done

chekunkov:
@vshlapakov tests are failing

vshlapakov:
@chekunkov AppVeyor build succeeded after rebase/push, but Travis gets stuck on OSX build, I guess it's related with [this issue](https://www.traviscistatus.com/incidents/k79mjcv403c4), I'll check it again a bit later.

dangra:
I found a problem related to this issue today while building a pure ruby image without `pip` installed.

Does it makes sense to check for specific python packages or should we backoff to check the kumo contract which is the common denominator for all images?

vshlapakov:
@dangra You're right, different programming languages lead to different sets of checks. Ok, as we're moving toward non-Scrapy crawling + we already have checks for main requirements of kumo contract (existence of `list-spiders` and `start-crawl` entries) - I think it's fine just to drop the check for sh-ep-scrapy.

chekunkov:
@vshlapakov @dangra check for sh_scrapy is useful for *many* project that used custom images for their Scrapy projects due to Scrapy Cloud limitations. not sure that removing sh_scrapy check is practical - if during conversion user will forget to install sh_scrapy his project will be broken and he'll barely understand why - that's why the check was originally added.

vshlapakov:
@chekunkov True, but we still check for `start-crawl` and `list-spiders` - that's my point. If during conversion user will forget to add sh_scrapy to requirements, there will be a corresponding [warning](https://github.com/scrapinghub/shub/blob/master/shub/image/test.py#L94). And if user implemented the entrypoints by himself, and they are searchable with PATH - I'd guess that he knows what he does and we'll only obstruct him with this check like in the case of the pure ruby spider.

My hopes are that it's clear enough from the docs that if client wants to port existing Scrapy crawler - he doesn't need to implement the protocol and use already existing library.

chekunkov:
got it, good point

shub/image/test.py
```diff
@@ -25,7 +25,7 @@
 """
 
 SH_EP_SCRAPY_WARNING = \
```
chekunkov: This is not needed anymore

vshlapakov: Wonder if we should keep it but reword somehow, like: 
```Command X is not found. Please make sure you provided it according to Scrapy Cloud contract (or added scrapinghub-entrypoint-scrapy(>=0.8.0) to your requirements file if you use Scrapy).```

Comment (chekunkov) [inline shub/image/test.py]: @vshlapakov sounds good to me

Comment (chekunkov) [inline shub/image/test.py]: worth adding a link to Scrapy Cloud contract, and remove `()` around `>=0.8.0`

Comment (vshlapakov) [inline shub/image/test.py]: Ok, fixed!

File: shub/image/test.py [hunk 3]
```diff
@@ -12,22 +12,20 @@
...
-SH_EP_SCRAPY_WARNING = \
-    'You should add scrapinghub-entrypoint-scrapy(>=0.7.0) dependency' \
-    ' to your requirements.txt or to Dockerfile to run the image with' \
-    ' Scrapy Cloud.'
+CONTRACT_CMD_NOT_FOUND_WARNING = \
+    'Command %s is not found in the image. ' \
+    'Please make sure you provided it according to Scrapy Cloud contract ' \
+    '(https://shub.readthedocs.io/en/stable/custom-images-contract.html) ' \
+    'or added scrapinghub-entrypoint-scrapy>=0.8.0 to your requirements ' \
+    'file if you use Scrapy.'
```
chekunkov: > Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

https://www.python.org/dev/peps/pep-0008/#maximum-line-length

pep8fied 😎

vshlapakov: Oh, of course, my bad!

chekunkov:
@jdemaeyer could you please take a look at and merge this PR?

jdemaeyer:
LGTM. I'll merge and release when I'm back from vacation on Monday

jdemaeyer:
Thanks viktor!

PR #231, opened 2017-02-13
Enable tests for Python 3.6

jdemaeyer:
Plus a small fix for `shub image check`: Previously, when updating the temporary file with status urls in `shub image check` once the line limit for that file had been reached, an _arbitrary_ line was removed via `dict.popitem()`. Because of an implementation detail in CPython, this used to fall in line with what we wanted, deleting the oldest entry. The fix makes sure we always explicitly delete the oldest entry. (The `dict` implementation has changed in CPython 3.6 and would now delete the newest dictionary entry instead.)

### DIFF
--- a/shub/image/utils.py
+++ b/shub/image/utils.py
@@ -145,7 +145,7 @@ def store_status_url(status_url, limit):
     status_id = max(data.keys()) + 1
     data[status_id] = status_url
     if len(data) > limit:
-        data.popitem()
+        del data[min(data.keys())]
     _update_status_file(data, STATUS_FILE_LOCATION)
     return status_id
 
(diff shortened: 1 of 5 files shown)


shub/image/utils.py
```diff
@@ -145,7 +145,7 @@ def store_status_url(status_url, limit):
     status_id = max(data.keys()) + 1
     data[status_id] = status_url
     if len(data) > limit:
-        data.popitem()
+        del data[min(data.keys())]
```
jdemaeyer: @vshlapakov Can you confirm that this is alright?

jdemaeyer: (see PR description)

vshlapakov: Confirming, you got the idea, thanks for fixing this!

Issue #228, opened 2017-01-13
shub image commands should be more consistent with other shub commands


chekunkov:
e.g. `shub deploy` supports `-V, --version TEXT` for version, while `shub image deploy` supports only `--version TEXT`

`shub image` commands use debug `-d` mode to enable verbose output while other shub commands have `-v` for that.

jdemaeyer:
Closed via #232

PR #232, opened 2017-02-13
Fix inconsistencies b/w shub & shub.image

vshlapakov:
Addressing #228.

Review, please. // cc @chekunkov


vshlapakov:
@jdemaeyer I think it should be fine to rename parameters as a part of migration process, for me "it's better now than later" mainly because of current ambiguity of `-d` parameter.

chekunkov:
I think it'd be better to keep backward compatibility for a couple of versions, print a warning if -d is used, drop -d (or change its meaning) after a couple of (or next) versions. This option is used mainly for debugging failed deploys, I think most of the users will notice the warning and migrate.

vshlapakov:
@chekunkov Ok, let's do it then, thanks for the comment!

chekunkov:
@vshlapakov please make sure there are no other discrepancies in the options, I mentioned only 2 the most obvious ones in the issue description

vshlapakov:
@chekunkov Added a deprecation message when using `-d`/`--debug` for all the command, haven't found any other inconsistencies except for:`-r`/`--requirements-file` for `shub deploy-reqs` VS `--requirements` for `shub image init`.

jdemaeyer:
I think we should stick with `--requirements` for `shub image init`, it's a little more handy than `--requirements-file`, and we're going to drop `shub deploy-reqs` soon.

vshlapakov:
@jdemaeyer Mm, I updated `shub image init` recently to use `--requirements-file` param instead of `--requirements`: my main argument is that first one is maybe less handy but says more about input expectations - that program expects a file path instead of a list of deps separated by commas, like for `base-deps`/`deps` [params](https://shub.readthedocs.io/en/stable/deploy-custom-image.html#init) for the same `init` command. Also `init` command usually is used pretty rare only for initial creation of Dockerfile, I don't think we need shorter name for the parameter.

jdemaeyer:
I agree that `--requirements-file` would have been the better choice for the parameter name.

But since

1. `--requirements` has already been established in `shub-image`, and 
2. we have already released a version of `shub` that also uses `--requirements`, and
3. the only `shub` command that uses `--requirements-file` is deprecated, so that there is no "acute" inconsistency in parameter names,

I would argue that this is not the right time to rectify misnamings.

vshlapakov:
@jdemaeyer Ok, now I'm convinced, thanks for argumentation! Fixed.

jdemaeyer:
Thanks alot viktor! :)

PR #233, opened 2017-02-13
Unify calling subcommands

jdemaeyer:
We had a couple of different implementations and helpers for calling subprocesses floating around, this PR unifies them.

It also removes the "Retry on EINTR" functionality as that has been [included in CPython](https://www.python.org/dev/peps/pep-0475/).

### DIFF
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -1,11 +1,9 @@
 from __future__ import absolute_import
 import os
-import sys
 import glob
 import shutil
 import tempfile
 from six.moves.urllib.parse import urljoin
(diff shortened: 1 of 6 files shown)


jdemaeyer:
Do we need to add deprecation wrappers for our util functions?

jdemaeyer:
@vshlapakov Is there a place in `shub.image` where subprocesses are called that I overlooked?

vshlapakov:
@jdemaeyer Nope, I don't think there're any subprocesses called by `shub.image` - its logic is based mainly on using `requests` and `docker-py`.

PR #235, opened 2017-03-02
Update setup.py

chekunkov:
Pin scrapinghub (>=1.9.0) and six (>=1.7.0) versions.
Add Python 3.6 trove classifier.

It may be a good idea to have versions for other requirements pinned to some meaningful ranges, e.g. I'm not sure shub works well with all versions of docker-py or all versions of click.

Fixes https://github.com/scrapinghub/shub/issues/239


setup.py
```diff
@@ -19,8 +19,16 @@
...
     include_package_data=True,
     zip_safe=False,
-    install_requires=['click', 'pip', 'requests', 'PyYAML', 'scrapinghub',
-                      'six', 'docker-py', 'retrying'],
+    install_requires=[
+        'click',
+        'docker-py',
+        'pip',
+        'PyYAML',
+        'retrying'
```
redapple: missing comma `,`

chekunkov: oh, thanks!

redapple:
@chekunkov , it seems six>=1.7.0 is needed too. See https://github.com/scrapinghub/shub/issues/239

chekunkov:
@redapple added

jdemaeyer:
Thanks guys

PR #236, opened 2017-03-02
Rename 'shub-image' to 'shub image' in messages and docstrings

stummjr:
Some messages and docstrings still refer to the legacy `shub-image` tool, such as this message I got when deploying an image:

    You can check deploy results later with 'shub-image check --id 0'.

This PR fixes those.


jdemaeyer:
Thanks @stummjr!

PR #242, opened 2017-03-29
Update Job URLs to new style

jdemaeyer:
With `job/` in the URL

### DIFF
--- a/shub/schedule.py
+++ b/shub/schedule.py
@@ -56,7 +56,7 @@ def cli(spider, argument, set):
                               targetconf.apikey, spider, argument, set)
     watch_url = urljoin(
         targetconf.endpoint,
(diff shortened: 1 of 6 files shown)


shub/schedule.py
```diff
@@ -56,7 +56,7 @@ def cli(spider, argument, set):
                               targetconf.apikey, spider, argument, set)
     watch_url = urljoin(
         targetconf.endpoint,
-        '../p/{}/job/{}/{}'.format(*job_key.split('/')),
+        '../p/{}/{}/{}'.format(*job_key.split('/')),
```
dangra: is there a point to split jobkey?

jdemaeyer:
Ah that's right, not anymore

PR #243, opened 2017-03-29
Test image before `shub image push`

chekunkov:
- Test image before `shub image push`.
- Skip tests if necessary with `-S/--skip-tests`.
- In `shub image upload` command tests are running during the build.
- Update tests for build, push and upload commands.
- Migrate test cases for build and push commands from unittest.TestCase to idiomatic pytest test cases with fixtures.
- Added new tests for -S/--skip-tests flag.
- Added project_dir fixture and removed boilerplate code from build and push commands tests.

### DIFF
--- a/shub/image/build.py
+++ b/shub/image/build.py
@@ -9,7 +9,7 @@ from shub.deploy import _create_default_setup_py
 from shub.utils import closest_file, get_config
 from shub.config import load_shub_config
 from shub.image import utils
-from shub.image import test as test_mod
+from shub.image import test as test_cmd
 
 
 SHORT_HELP = 'Build release image.'
@@ -67,7 +67,7 @@ def build_cmd(target, version, skip_tests):
     click.echo("The image {} build is completed.".format(image_name))
     # Test the image content after building it
     if not skip_tests:
-        test_mod.test_cmd(target, version)
+        test_cmd.test_cmd(target, version)
 
 
 def _create_setup_py_if_not_exists():
--- a/shub/image/push.py
+++ b/shub/image/push.py
@@ -31,12 +31,12 @@ otherwise you have to enter your credentials (at least username/password).
 @click.option("-v", "--verbose", is_flag=True,
               help="stream push logs to console")
 @click.option("-V", "--version", help="release version")
-@click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
 @click.option("--username", help="docker registry name")
 @click.option("--password", help="docker registry password")
 @click.option("--email", help="docker registry email")
 @click.option("--apikey", help="SH apikey to use built-in registry")
 @click.option("--insecure", is_flag=True, help="use insecure registry")
+@click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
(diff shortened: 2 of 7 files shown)


vshlapakov:
@chekunkov Looks good, nice idea. So now if user ignores test errors (or skips the tests) on build step, push step will fail too with same errors if user didn't fix anything, that's the point?

shub/image/build.py
```diff
@@ -67,7 +67,7 @@ def build_cmd(target, version, skip_tests):
     click.echo("The image {} build is completed.".format(image_name))
     # Test the image content after building it
     if not skip_tests:
-        test_mod.test_cmd(target, version)
+        test_cmd.test_cmd(target, version)
```
vshlapakov: Do you think that now it looks better for some reasons? Just out of curiosity, I'm trying to understand motivation as I'd rename it rather to full `test_module`. I was always a bit confused with naming like `from datetime import datetime` 🙂

shub/image/push.py
```diff
@@ -30,17 +31,22 @@
 @click.option("-v", "--verbose", is_flag=True,
               help="stream push logs to console")
 @click.option("-V", "--version", help="release version")
+@click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
```
vshlapakov: Let's move this line to the end of the section, to match its order with function args order for better readability?

shub/image/build.py
chekunkov: test_module can be anything, test_cmd is a test command, I think it's just easier to digests.

chekunkov:
@vshlapakov yes, you nailed it.

shub/image/build.py
vshlapakov: 👌

chekunkov:
@jdemaeyer could you please review and merge this? I have another branch where I can reuse some of the new fixtures I've added here, don't want to copy-paste the code.

shub/image/build.py
jdemaeyer: I'll merge this with gritted teeth, but my pedantic streak can't let it pass without commenting. ;)

A command is very narrowly defined thing specified by `click`. It must be an instance of `click.Command` (or at least behave like it). The `@click.command` decorator "creates a new `Command` and uses the decorated function as callback" . A module is a different very narrowly defined thing specified by the Python language.

Just because a module contains a `Command` at its heart (_among_ a bunch of helper functions) it doesn't suddenly become a command, and personally I find `test_cmd.test_cmd()` more difficult to digest than `test_mod.test_cmd()`. Why would a command live within another command of the same name?

```python
In [1]: from shub.image import test as test_cmd

In [2]: type(test_cmd)
Out[2]: module

In [3]: type(test_cmd.cli)
Out[3]: click.core.Command
```

chekunkov: ahaha, ok @jdemaeyer @vshlapakov, I'll fix this :) 2 options: 

1. I can rename it back 
2. import command only

```
from shub.image.test import test_cmd
# ...
test_cmd(target, version)
```
I'd go with the last one

jdemaeyer:
I like the last one too =)

PR #244, opened 2017-03-30
Add singleton configuration

jdemaeyer:
shub's syntax for advanced use cases like setting a custom stack is pretty verbose and to many users confusing. This PR adds singletons to simplify setting the `default` key for dictionary-like options (`projects`, `endpoints`, `apikeys`, `stacks`, `images`). I.e., this configuration:

```
projects:
    default: 123
stacks:
    default: scrapy:1.3-py3
```

can now also be achieved via

```
project: 123
stack: scrapy:1.3-py3
```

Thanks @dangra for the idea :)


dangra:
LGTM!

jdemaeyer:
Thanks for the review Daniel! 

@chekunkov you also have a good eye for user experience. You think we should add anything while we're at it (or change)?

chekunkov:
@jdemaeyer nothing comes to mind - nice idea, everything looks good to me

PR #246, opened 2017-03-30
Fix shub image init

chekunkov:
It failed with TypeError if there was not scrapy.cfg in the current directory.

Update help message and error message.
Update base image in Dockerfile template.
Update list of recommended Python packages.
Use click.confirm to prompt for save file confirmation.
Deprecate --base-deps parameter.
Update init command tests.

### DIFF
--- a/shub/image/init.py
+++ b/shub/image/init.py
@@ -3,7 +3,6 @@ import textwrap
 from string import Template
 
 import click
-from six.moves import input
 
 from shub import exceptions as shub_exceptions
 from shub import utils as shub_utils
@@ -11,7 +10,7 @@ from shub.image import utils
 
 
 DOCKER_APP_DIR = '/app'
-DOCKERFILE_TEMPLATE = """
+DOCKERFILE_TEMPLATE = """\
 FROM $base_image
 $system_deps
 $system_env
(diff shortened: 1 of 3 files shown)


chekunkov:
BTW @vshlapakov @redapple it seems like this command is only suitable for the migration of Scrapy projects. I'll update documentation and mention that init step is only necessary for the users who already have existing Scrapy project in a separate PR, because there are other changes I want to add to custom image deploy docs.

chekunkov:
Also @vshlapakov wdyt if we start using scrapinghub-stack-scrapy:1.3 as a base? This will allow to remove half of the Python packages and all system packages from init command.

vshlapakov:
I would say it's fair enough: we can't predict what user is going to achieve for sure, so our best guess is that it will be a custom image with Scrapy-based crawler, and in this case using last stable stack is perfect.

chekunkov:
Ready for review. I've tested with the oldest Scrapy project I was able to find - the generated Dockerfile worked just fine and spider finished successfully.

shub/image/init.py
```diff
@@ -22,42 +21,42 @@
...
-via --requirements option, we'll create new requirements.txt in the
-project_dir with the recommended deps(use --list-recommended-reqs to list it).
+System packages
 
-You should also include scrapinghub-entrypoint-scrapy package to your py-reqs,
-it's a necessary condition to run your project with Kumo (use 'shub image test'
-to check if your image fits our expectations).
-""".format(BASE_SYSTEM_DEPS)
+You can extend list of system packages installed in the image via --add-deps option or
+provide a comma-separated list of packages via --base-deps option.
```
jdemaeyer: `--add-deps` is also a comma-separated list, right? Let's remove `--base-deps` from the help message if we deprecated it.

chekunkov: good point, fixed

vshlapakov:
LGTM 👍

shub/image/init.py
```diff
@@ -98,23 +107,16 @@ def cli(project, base_image, base_deps, add_deps, requirements):
...
-        choice = input("Save to {}: (y/n)".format(dockefile_path)).lower()
-        if choice in valid:
-            if valid[choice]:
-                with open(dockefile_path, 'w') as dockerfile:
-                    dockerfile.write(results)
-                click.echo('Saved.')
-            break
-        click.echo("Please respond with 'yes'('y') or 'no'(n)")
+    dockefile_path = os.path.join(project_dir, 'Dockerfile')
+    if click.confirm("Save to {}?".format(dockefile_path)):
```
vshlapakov: To discuss for future: I wonder if this wizard is required at all, probably it's better to simply create Dockerfile or fail with a warning if it already exists, wdyt?

shub/image/init.py
```diff
@@ -70,26 +68,37 @@ def list_recommended_python_reqs(ctx, param, value):
...
+
+
 @click.command(help=HELP, short_help=SHORT_HELP)
 @click.option("--list-recommended-reqs", is_flag=True, is_eager=True,
               expose_value=False, callback=list_recommended_python_reqs,
               help="list recommended python requirements")
 @click.option("--project", default="default",
               help="project name to get settings module from scrapy.cfg")
-@click.option("--base-image", default="python:2.7",
+@click.option("--base-image", default="scrapinghub/scrapinghub-stack-scrapy:1.3",
```
vshlapakov: I'd probably move the base image to some constant `DEFAULT_BASE_IMAGE` on top of the module, but that's a minor thing and can be done later.

shub/image/init.py
```diff
@@ -22,42 +21,41 @@
...
 BASE_PYTHON_DEPS = [
-    'BeautifulSoup==3.2.0', 'MySQL-python==1.2.3', 'Scrapy==1.0.3',
-    'Twisted==11.1.0', 'boto==2.28.0', 'guppy==0.1.10', 'hubstorage==0.19.1',
-    'ipython==3.2.1', 'lxml==3.4.4', 'numpy==1.6.1', 'pika==0.9.14',
-    'psycopg2==2.4.5', 'pymongo==2.6.3', 'queuelib==1.2.2',
-    'raven==5.0.0', 'requests==2.5.3', 'scrapely==0.12.0',
-    'scrapinghub==1.7.0', 'scrapylib==1.6.0', 'setproctitle==1.0.1',
-    'slybot==0.12.1', 'w3lib==1.13.0',
-    'scrapinghub-entrypoint-scrapy==0.6.0'
+    'BeautifulSoup==3.2.0',
```
vshlapakov: The requirements were outdated mainly because it's hard-coded in the sources. Wdyt if we move it to some `base-requirements.txt` and drop a reminder to update it from time to time to README? Though it strongly correlates with chosen base image,  having some doubts if it improves the process and not complicates it.

vshlapakov:
@jdemaeyer I left some notes, but it's mainly for discussion; if you don't mind, I would merge this PR, it looks ready and there's another pending PR built on top of the changes.

shub/image/init.py
jdemaeyer: Good point. Since there's no configuration questions being asked before the Dockerfile is generated there's no use in having a "dry run" functionality. Just fail directly at the beginning when `Dockerfile` already exists, otherwise write it without extra confirmation.

shub/image/init.py
jdemaeyer: This is probably because I haven't worked much with custom images from a user perspective yet, but I have a little trouble understanding where these recommended dependencies come from? If they are required for Scrapy or Scrapy Cloud then then they should be part of `scrapinghub-entrypoint-scrapy` or `stack-scrapy:1.3`.

IPython and Guppy I reckon are for debugging inside the container which I can somehow imagine happening. But why are we adding Portia, NumPy, and the database/rabbitmq adapters by default? If the user uses these in a pipeline/middleware they will have them in their own `requirements.txt` for sure. In particular `numpy` is a 70 MB package once unpacked.

jdemaeyer:
Thanks for reviewing again @vshlapakov. I wouldn't mind to have this open until we've settled the discussions, but if you think it would speed other things up feel free to merge already.

shub/image/init.py
chekunkov: Yeah, I have the same concerns, we can actually remove these recommended python packages, `scrapy:1.3` has everything needed to run Scrapy spiders, other packages can be manually added by users.

chekunkov: Or, as a less radical suggestion, we can limit list of recommended python packages to guppy :)

vshlapakov: Clean it by keeping only `guppy` sounds good to me, it could be useful (especially for us) to debug different memory issues. Most of the other packages were added long time ago to match with `hworker` stack that was default one at that time, there's no real need to keep it for now.

jdemaeyer: Sounds fair

PR #247, opened 2017-03-31
Update shub image test command imports

chekunkov:
Updated test fixtures because imported command should be mocked in each individual module.
Also sorted imports in the shub.image.build and shub.image.pull modules.


jdemaeyer:
Thanks Alex!

PR #248, opened 2017-04-03
Shub image push progress with tqdm

vshlapakov:
The goal of the changes is to improve verbosity of push progress, cause events stream in verbose mode is too verbose and hard to read, and non-verbose mode is silent and can wait for a lot of time w/o any sign of life, especially when there're a lot of layers to push and network is flaky.

Initial attempt was to achieve similar results with built-in [click.progressbar](http://click.pocoo.org/5/api/#click.progressbar), but there're a few issues with it:
- it doesn't support multiple bars and it's tricky to work it around
- single bar works fine, but we don't have any variable to track total progress with enough details - only as amount of layers left to push (but if some layers are large it doesn't work good)
- almost always several layers are being pushed at the same time, and it's hard to represent its progress properly

As a result, I wasn't satisfied with built-in `click.progressbar` possibilities and decided to try [tqdm](https://github.com/tqdm/tqdm) library as it's much more flexible, provides multiple bars out-of-the-box and pretty configurable (also it's pretty lightweight and should work fine everywhere).

The proposed workflow also brings light on some complexities related with handling docker-push events stream:
- it's not straightforward to estimate amount of layers to push - `docker inspect` is closest approach, but it works differently on different platform and thus not very reliable (on OSX layers are in `RootFS->Layers`, and on Linux layers are not showed this way)
- we don't know size of pushed layer in advance because docker compresses it before sending

Please, review. // cc @chekunkov

### DIFF
--- a/shub/image/push.py
+++ b/shub/image/push.py
@@ -1,3 +1,6 @@
+import sys
+from collections import OrderedDict
+
 import click
 
 from shub import exceptions as shub_exceptions
@@ -58,16 +61,15 @@ def push_cmd(target, version, username, password, email, apikey, insecure, skip_
         _execute_push_login(client, image, username, password, email)
     image_name = utils.format_image_name(image, version)
     click.echo("Pushing {} to the registry.".format(image_name))
-    for data in client.push(image_name, stream=True, decode=True,
-                            insecure_registry=not bool(username)):
-        if 'status' in data:
-            utils.debug_log("Logs:{} {}".format(data['status'],
-                            data.get('progress')))
-        if 'error' in data:
-            click.echo("Error {}: {}".format(data['error'],
-                                             data['errorDetail']))
-            raise shub_exceptions.RemoteErrorException(
-                "Docker push operation failed")
+    events = client.push(image_name, stream=True, decode=True,
+                         insecure_registry=not bool(username))
+    ctx = click.get_current_context(True)
+    if ctx and ctx.params.get('verbose'):
+        push_progress_cls = _LoggedPushProgress
+    else:
+        push_progress_cls = _PushProgress
+    push_progress = push_progress_cls(events)
+    push_progress.show()
     click.echo("The image {} pushed successfully.".format(image_name))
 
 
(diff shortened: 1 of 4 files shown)


chekunkov:
Doesn't work that well for me, bars are duplicated mid way

```Pushing images.scrapinghub.com/project/23661:r11-default to the registry.
total:   0%|                                                                                 | 0/20
total:   5%|████                                                                             | 1/20
cd1037afabaa: 100%|███████████████████████████████████████████| 24.8K/24.8K [00:06<00:00, 3.63KB/s]
/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/config.py:246: UserWarning: get_target is deprecated, use get_target_conf instead
```

chekunkov:
Another thing to note - push is not the only operation that can hang for a long time. Build can, by default we provide no output there as well. Deploy can, and again, no output. We need to address those issues as well to close the ticket.

chekunkov:
check https://github.com/click-contrib/click-spinner

vshlapakov:
@chekunkov Weird, I tested it a lot - bars are duplicated for me only if I press enter while the process (not sure if it's a problem though, Docker behaves in the same way), or it's duplicated even if you don't press anything? You're using OSX, right?

The solution with tqdm looks a way better than click.progressbar or any spinner, but it should work like a good clock, I agree that it's important.

Using click-spinner looks easy, but not sure if it gives any useful information to user, except that process is working on something, is this all that what we need for now?

chekunkov:
@vshlapakov don't press anything, I'm using OSX, correct. I'll try to debug the issue later today.

something like click-spinner can be used for `shub image deploy`, where we don't really have a progress bar - as I stressed above this is another time consuming operations where user doesn't have any feedback.

chekunkov:
@vshlapakov I can reproduce this issue almost each time I do a build followed by a push, but for several subsequent pushes it looks okay, can be the reason why you don't see that issue.

For me docker push produced the following [events](https://github.com/scrapinghub/shub/files/891090/debug.txt) for the output

```
→ shub image build && shub image push
The image images.scrapinghub.com/project/23661:r11-default build is completed.
Login to images.scrapinghub.com succeeded.
Pushing images.scrapinghub.com/project/23661:r11-default to the registry.
total:   0%|                                                                                  | 0/20
203225d657f3:   0%|                                                     | 0.00/24.8K [00:00<?, ?B/s]
1c01ff6502b6: 100%|████████████████████████████████████████████| 26.3K/26.3K [00:06<00:00, 4.00KB/s]
```

Does it help?

shub/image/push.py
```diff
@@ -177,3 +155,33 @@ def show(self):
...
+def _create_total_bar():
+    return utils.ProgressBar(
+        total=1,
+        desc='total',
+        file=click.get_text_stream('stdout'),
+        dynamic_ncols=True,
+        miniters=1,
+        # don't need rate here, let's simplify the bar
+        bar_format='{l_bar}{bar}| {n_fmt}/{total_fmt}'
+    )
```
chekunkov: I think these should be methods in the class - in case someone would like to change the behavior and override them

jdemaeyer:
LGTM 👍 , definitely squash before merging ;)

PR #249, opened 2017-04-03
Update Deploying custom Docker images documentation

chekunkov:
please review @vshlapakov @redapple

### DIFF
--- a/docs/custom-images-contract.rst
+++ b/docs/custom-images-contract.rst
@@ -4,8 +4,8 @@
 Custom Images contract
 ======================
 
-A contract is a set of requirements that any crawler custom `Docker`_ image have to comply
-with to be able to run on Scrapy Cloud.
+This is a set of requirements that any custom `Docker`_ image has to comply with
+to be able to run on Scrapy Cloud.
 
 Scrapy crawler Docker images are already supported via
 the :ref:`scrapinghub-entrypoint-scrapy <sh-entrypoint-scrapy>` contract implementation.
--- a/docs/deploy-custom-image.rst
+++ b/docs/deploy-custom-image.rst
@@ -4,188 +4,173 @@
 Deploying custom Docker images
 ==============================
 
-Scrapy Cloud 2.0 allows you to run your spiders in custom Docker containers. The tools for managing and deploying Docker images are bundled under ``shub image``.
+.. note::
+
+    This feature is currently only available for paying customers.
+
+It's possible to deploy Docker images with spiders to Scrapy Cloud. To be able to run spiders in custom `Docker`_
+images it's necessary to follow the :ref:`Custom images contract <custom-images-contract>` - a set of requirements
+that image should comply with to be compatible with Scrapy Cloud.
+
+.. _Docker: https://docs.docker.com/
 
 Deployment
 ==========
-This section describes how to build and deploy to Scrapy Cloud 2.0 a custom Docker image for a Scrapy project.
 
-.. _step-one:
+This section describes how to build and deploy a custom Docker image to Scrapy Cloud. For all the following steps
+it's assumed that commands are executed at the root directory of your project.
 
-1. Initialization
------------------
-Open up a terminal and go to your crawler's project folder in your local machine::
+1. Create Dockerfile
+--------------------
 
-    $ cd path/to/your/project
+The most important thing you need to be able to build and deploy Docker images is a `Dockerfile`_.
+Please follow the link if you are not familiar with the concept as it's crucial to understand it
+while using custom Docker images feature.
 
-And then run the :ref:`init command <commands-init>`::
+If you want to migrate an existing Scrapy project - there's a tool that may help you, please read
+:ref:`this section <create-image-for-scrapy-project>`. In all other cases you're responsible for writing your own
+Dockerfile. The resulting Dockerfile should produce a Docker image that follows the
+:ref:`Custom images contract <custom-images-contract>` - follow the link to find an example Dockerfile.
 
-    $ shub image init --requirements path/to/requirements.txt
+.. _Dockerfile: https://docs.docker.com/engine/reference/builder/
 
-This will create a Dockerfile for your container including ``requirements.txt`` as a dependency and using `python:2.7 <https://hub.docker.com/r/library/python/>`_ as `the base image <https://docs.docker.com/engine/reference/builder/>`_ for your custom image. If you want to use a different one, pass the ``--base-image`` option, like this::
+2. Deploy to Scrapy Cloud
+-------------------------
 
-    $ shub image init --base-image scrapinghub/base:12.04
+Once you have the Dockerfile run the :ref:`shub deploy <basic-usage>` command to build the Docker image.
+If there's no :ref:`scrapinghub.yml <configuration>` configuration file at the project root
+shub will start a wizard that will help to configure the project and will save the configuration file.
+If you already have :ref:`scrapinghub.yml <configuration>` at the project root please ensure that
+:ref:`image deploy is configured <configuration-options>` for the target project. If the target project
+already exists in the configuration file but images deploy is not configured you can run
+:ref:`shub image build <commands-upload>` to build the image for the first time and shub
+will help you to configure the image repository.
 
-In this case, it will use the image available at https://hub.docker.com/r/scrapinghub/base tagged with ``12.04``.
+The deploy consists of 3 stages which are described below. Normally :ref:`shub deploy <basic-usage>` will execute
+all 3 stages in a single run, but in some cases in might be useful to run those stages separately,
+so there are commands bundled under :ref:`shub image <commands>` that allow to execute different stages separately.
 
-.. warning:: If you want to use Scrapy in the custom Docker image you may want to include
-    `scrapinghub-entrypoint-scrapy`_ in your requirements.txt file.
-    It is a support layer that passes data from the job to Scrapinghub storage.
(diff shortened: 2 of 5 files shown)


docs/deploy-custom-image.rst
```diff
@@ -4,90 +4,114 @@
...
 
-In this case, it will use the image available at https://hub.docker.com/r/scrapinghub/base tagged with ``12.04``.
+1. Add scrapinghub.yml
+----------------------
 
-.. warning:: If you want to use Scrapy in the custom Docker image you may want to include
-    `scrapinghub-entrypoint-scrapy`_ in your requirements.txt file.
-    It is a support layer that passes data from the job to Scrapinghub storage.
-    Otherwise you will need to send data to Scrapinghub storage using `HTTP API`__.
+:ref:`scrapinghub.yml <configuration>` is a Scrapinghub configuration file that defines deploy targets
```
vshlapakov: It's probably not clear what are the `deploy targets`, maybe just drop this part?
```
is a Scrapinghub configuration file that links local code repository with projects
```

docs/deploy-custom-image.rst
```diff
@@ -4,90 +4,114 @@
...
 
-This will create a Dockerfile for your container including ``requirements.txt`` as a dependency and using `python:2.7 <https://hub.docker.com/r/library/python/>`_ as `the base image <https://docs.docker.com/engine/reference/builder/>`_ for your custom image. If you want to use a different one, pass the ``--base-image`` option, like this::
+This section describes how to build and deploy a custom Docker image to Scrapy Cloud. For all the following steps
+it's assumed that commands are executed at the root directory of your project.
 
-    $ shub image init --base-image scrapinghub/base:12.04
 
-In this case, it will use the image available at https://hub.docker.com/r/scrapinghub/base tagged with ``12.04``.
+1. Add scrapinghub.yml
+----------------------
```
jdemaeyer: After #252 we should probably add a note here that this step can be skipped and a wizard will appear during step three Wizard
Or (more drastically) move it to a separate section like the `shub image init` info

vshlapakov: I'm afraid the guide should be revised again after both #252 and #253, last one updates recommended changes in the configuration file + introduces `shub deploy` working with custom images as a shortcut to `shub image upload`

Nice wizard btw 😄

chekunkov: Sure, that's why it's moved it to WIP, I'll review docs after ^ major changes are merged

chekunkov: @jdemaeyer @vshlapakov I've reviewed this doc once again - it seems to me that this sections should be removed completely if shub takes care of the config. Instead I'll add a note to the `Build the image` step that will mention the wizard. Please let me know if any info in this section is valuable - we can move it to https://shub.readthedocs.io/en/stable/configuration.html

Any pushback?

vshlapakov: Agree, we anyway have to describe last changes for shub config. I think mentioning the wizard and putting a link to shub config documentation for advanced cases is more than enough

jdemaeyer: I agree that we can remove this section. Since @dangra suggested that we drop support for custom repositories completely at some point, I wonder if we should simply not mention it in the docs at all?

The https://shub.readthedocs.io/en/stable/configuration.html needs some love, it would be good to have a "full-fledged" configuration in there and have a list of all available options. I can prepare a PR for that tomorrow

chekunkov: > I wonder if we should simply not mention it in the docs at all?
I'm okay with that

> I can prepare a PR for that tomorrow
Thanks!

chekunkov: Talking about support for custom repositories. There's one use case we still need to support I think - for local development we may want to replace images.scrapinghub.com with another host:port? Do you have any good idea how this can/will be implemented if we drop custom repos?

jdemaeyer: Hmm good question. Maybe we can never completely drop support for custom repos, but we can treat them as undocumented/private

chekunkov:
@vshlapakov @jdemaeyer please review the updated PR. I'll squash everything in a single commit before merge.

docs/custom-images-contract.rst
```diff
@@ -4,8 +4,8 @@
 Custom Images contract
 ======================
 
-A contract is a set of requirements that any crawler custom `Docker`_ image have to comply
-with to be able to run on Scrapy Cloud.
+This is a set of requirements that any custom `Docker`_ image have to comply with
```
jdemaeyer: While we're at it, could you replace `have` -> `has`?

docs/deploy-custom-image.rst
```diff
@@ -4,175 +4,151 @@
...
+-------------------
 
-The settings above define that shub will push the image of your Docker container to https://hub.docker.com/r/yourusername/repository. You can also specify the complete URL for your repository if you are not using the default registry (which is https://hub.docker.com).
+The most important thing you need to be able to build and deploy Docker images is a `Dockerfile`_.
+Please follow the link if you are not familiar with the concept as it's crucial to understand it
+while using custom Docker images feature.
 
-.. tip:: Your project might not have a ``scrapinghub.yml`` file, because it has been introduced with recent versions of shub. Make sure to upgrade your shub package by running::
+If you want to migrate an existing Scrapy project - there's a tool that may help you, please read
+`this section <create-image-for-scrapy-project>`_. In all other cases you're responsible for writing your own
```
jdemaeyer: This link is not working for me. I think you need to use `:ref:`

docs/deploy-custom-image.rst
```diff
@@ -4,175 +4,151 @@
...
+Deployment
+==========
 
-    projects:
-        default: 29629
-    image: yourusername/repository
+This section describes how to build and deploy a custom Docker image to Scrapy Cloud. For all the following steps
+it's assumed that commands are executed at the root directory of your project.
 
+1. Create Dockefile
```
jdemaeyer: Dockerfile

docs/deploy-custom-image.rst
```diff
@@ -4,175 +4,151 @@
...
-The settings above define that shub will push the image of your Docker container to https://hub.docker.com/r/yourusername/repository. You can also specify the complete URL for your repository if you are not using the default registry (which is https://hub.docker.com).
+The most important thing you need to be able to build and deploy Docker images is a `Dockerfile`_.
+Please follow the link if you are not familiar with the concept as it's crucial to understand it
+while using custom Docker images feature.
 
-.. tip:: Your project might not have a ``scrapinghub.yml`` file, because it has been introduced with recent versions of shub. Make sure to upgrade your shub package by running::
+If you want to migrate an existing Scrapy project - there's a tool that may help you, please read
+`this section <create-image-for-scrapy-project>`_. In all other cases you're responsible for writing your own
+Dockerfile. The resulting Dockerfile should produce a Docker image that follows the
+:ref:`Custom images contract <custom-images-contract>` - follow the link to find an example Dockefile.
```
jdemaeyer: Dockerfile :)

docs/deploy-custom-image.rst
```diff
@@ -4,175 +4,151 @@
...
-    And then create ``scrapinghub.yml`` by running::
+2. Deploy to Scrapy Cloud
+-------------------------
 
-            $ shub deploy
+Once you have the Dockerfile run the :ref:`shub deploy <basic-usage>` command to build the Docker image.
+If there's no :ref:`scrapinghub.yml <configuration>` configuration file at the project root
+shub will start a wizard that will help to configure the project and will save the configuration file.
+If you already have :ref:`scrapinghub.yml <configuration>` at the project root please ensure that
+:ref:`image deploy is configured <configuration-options>` for the target project.
```
jdemaeyer: For typical use cases, there is an easier way to configure image deploy even if there is a `scrapinghub.yml`: `shub image upload` will run the "configure repository" part of the wizard. Maybe we can mention that?

docs/deploy-custom-image.rst
```diff
@@ -4,175 +4,151 @@
...
 
 build
 -----
-This command uses the Dockerfile created by the :ref:`init <commands-init>` command to build the image that's going to be deployed later.
 
-It reads the target images from the `scrapinghub.yml <http://doc.scrapinghub.com/shub.html#configuration>`_ file, which is generated by the deploy command from shub >= 2.0. You should add a section called ``images`` on it using the following format:
+This command uses the Dockerfile to build the image that's going to be deployed later.
+
+It reads the target images from the :ref:`scrapinghub.yml <configuration>` file.
+You should add a section called ``images`` on it using the following format:
```
jdemaeyer: The example below this uses the deprecated `images` section

docs/deploy-custom-image.rst
chekunkov: I'm a bit reluctant to advertize `shub image upload` because I hope that eventually we will remove it and keep only `shub deploy`, but I'm okay to add this for now.

chekunkov:
Thanks for proofreading @jdemaeyer, fixed

docs/deploy-custom-image.rst
jdemaeyer: Hmm I see, that makes sense. Maybe we can advertise `shub image build` instead?

chekunkov: @jdemaeyer updated, does it look better now?

jdemaeyer: lgtm!

PR #251, opened 2017-04-05
Progress bar for shub image build operation

vshlapakov:
This one relies on native `click.progressbar` to provide a bar for `shub image build`, please review.

### DIFF
--- a/shub/image/build.py
+++ b/shub/image/build.py
@@ -24,6 +24,9 @@ provided, the tool uses VCS-based stamp over project directory (the same as
 shub utils itself).
 """
 
+BUILD_STEP_REGEX = re.compile(r'Step (\d+)/(\d+) :.*')
+BUILD_SUCCESS_REGEX = re.compile(r'Successfully built ([0-9a-f]+)')
+
 
 @click.command(help=HELP, short_help=SHORT_HELP)
 @click.argument("target", required=False, default="default")
@@ -49,19 +52,67 @@ def build_cmd(target, version, skip_tests):
     if not os.path.exists(os.path.join(project_dir, 'Dockerfile')):
         raise shub_exceptions.BadParameterException(
             "Dockerfile is not found, please use shub image 'init' command")
-    is_built = False
-    for data in client.build(path=project_dir, tag=image_name, decode=True):
-        if 'stream' in data:
-            utils.debug_log("{}".format(data['stream'][:-1]))
-            is_built = re.search(
-                r'Successfully built ([0-9a-f]+)', data['stream'])
-        elif 'error' in data:
-            click.echo("Error {}:\n{}".format(
-                data['error'], data['errorDetail']))
-    if not is_built:
-        raise shub_exceptions.RemoteErrorException(
-            "Build image operation failed")
+    if utils.is_verbose():
+        build_progress_cls = _LoggedBuildProgress
+    else:
+        build_progress_cls = _BuildProgress
+    click.echo("Building {}.".format(image_name))
+    events = client.build(path=project_dir, tag=image_name, decode=True)
+    build_progress = build_progress_cls(events)
+    build_progress.show()
     click.echo("The image {} build is completed.".format(image_name))
     # Test the image content after building it
     if not skip_tests:
         test_cmd(target, version)
+
+
+class _LoggedBuildProgress(utils.BaseProgress):
+    """Visualize build progress in verbose mode.
+
+    Output all the events received from the docker daemon.
+    """
+    def handle_event(self, event):
+        super(_LoggedBuildProgress, self).handle_event(event)
+        if 'stream' in event:
+            self.handle_stream_event(event)
+
+    def handle_stream_event(self, event):
+        utils.debug_log("{}".format(event['stream'][:-1]))
+
+
+class _BuildProgress(_LoggedBuildProgress):
+    """Visualize build progress in non-verbose mode.
+
+    Show total progress bar.
+    """
+
+    def __init__(self, events):
+        super(_BuildProgress, self).__init__(events)
+        self.bar = utils.create_progress_bar(
+            total=1,
+            desc='Steps',
+            # don't need rate here, let's simplify the bar
+            bar_format='{l_bar}{bar}| {n_fmt}/{total_fmt}',
+        )
+        self.is_built = False
+
+    def show(self):
+        super(_BuildProgress, self).show()
+        if self.bar:
+            self.bar.close()
+        if not self.is_built:
+            raise shub_exceptions.RemoteErrorException(
+                "Build image operation failed")
+
+    def handle_stream_event(self, event):
+        if BUILD_SUCCESS_REGEX.search(event['stream']):
+            self.is_built = True
+            return
+        step_row = BUILD_STEP_REGEX.match(event['stream'])
+        if not step_row:
+            return
+        step_id, total = [int(val) for val in step_row.groups()]
+        self.bar.total = max(self.bar.total, total)
+        # ignore onbuild sub-steps
+        if step_id > self.bar.n and self.bar.total == total:
+            self.bar.update()
--- a/shub/image/push.py
+++ b/shub/image/push.py
@@ -1,4 +1,3 @@
-import sys
 from collections import OrderedDict
 
 import click
@@ -63,8 +62,7 @@ def push_cmd(target, version, username, password, email, apikey, insecure, skip_
     click.echo("Pushing {} to the registry.".format(image_name))
     events = client.push(image_name, stream=True, decode=True,
                          insecure_registry=not bool(username))
-    ctx = click.get_current_context(True)
-    if ctx and ctx.params.get('verbose'):
+    if utils.is_verbose():
         push_progress_cls = _LoggedPushProgress
     else:
         push_progress_cls = _PushProgress
@@ -86,27 +84,15 @@ def _execute_push_login(client, image, username, password, email):
     click.echo("Login to {} succeeded.".format(registry))
 
 
-class _LoggedPushProgress(object):
+class _LoggedPushProgress(utils.BaseProgress):
     """Visualize push progress in verbose mode.
 
     Output all the events received from the docker daemon.
     """
-
-    def __init__(self, push_events):
-        self.push_events = push_events
-
-    def show(self):
-        for event in self.push_events:
-            self.handle_event(event)
-
     def handle_event(self, event):
+        super(_LoggedPushProgress, self).handle_event(event)
         if 'status' in event:
             self.handle_status_event(event)
-        if 'error' in event:
-            click.echo("Error {}: {}".format(event['error'],
-                                             event['errorDetail']))
-            raise shub_exceptions.RemoteErrorException(
-                "Docker push operation failed")
 
     def handle_status_event(self, event):
         msg = "Logs:{} {}".format(event['status'], event.get('progress'))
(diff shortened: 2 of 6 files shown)


shub/image/build.py
```diff
@@ -70,6 +75,26 @@ def build_cmd(target, version, skip_tests):
         test_mod.test_cmd(target, version)
 
 
+def _create_or_update_progressbar(bar, event, verbose):
+    """Helper to create progress bar to track progress."""
+    step_row = STEP_REGEX.match(event['stream'])
```
chekunkov: sometimes there are substeps, caused by build triggers, do you ignore them properly?

1__alex_mbp____dev_workspace_scrapinghub_proj_scrapy-first-project_hackershelf__zsh__and_progress_bar_for_shub_image_build_operation_by_vshlapakov_ _pull_request__251_ _scrapinghub_shub

shub/image/utils.py
```diff
@@ -31,10 +31,13 @@
 """
 
 
-def debug_log(msg):
+def is_verbose():
     ctx = click.get_current_context(True)
-    # FIXME remove check of debug param as it's deprecated
-    if ctx and (ctx.params.get('verbose') or ctx.params.get('debug')):
+    return ctx and ctx.params.get('verbose')
```
jdemaeyer: We can't remove the check for `debug` unless we also remove the `--debug` parameter from all the `shub image` commands (which in turn we can't do unless we release v3.0 because we follow semantic versioning).

We should either leave the `debug` check here or alternatively we can remove it and add a callback for the `--debug` parameters that sets `ctx['verbose'] = True`

shub/image/build.py
vshlapakov: Good catch, thanks!

shub/image/utils.py
vshlapakov: I removed it here because I had a feeling that using `debug` parameter is already disabled and doesn't work except for raising a warning: this [function](https://github.com/scrapinghub/shub/blob/68623011256b396954b278656f5a0ac919f3cb24/shub/image/utils.py#L40) doesn't return anything, so context doesn't contain `debug` parameter. If this is considered as a wrong behaviour, let's fix callback logic.

jdemaeyer: Uhh you're right, there is a bug in the callback, it should `return value`. Can you add it?

shub/image/build.py
```diff
@@ -53,15 +57,20 @@ def build_cmd(target, version, skip_tests):
...
     for data in client.build(path=project_dir, tag=image_name, decode=True):
         if 'stream' in data:
+            if not verbose:
+                bar = _create_or_update_progress_bar(bar, data, verbose)
             utils.debug_log("{}".format(data['stream'][:-1]))
             is_built = re.search(
                 r'Successfully built ([0-9a-f]+)', data['stream'])
         elif 'error' in data:
-            click.echo("Error {}:\n{}".format(
+            tqdm.write("Error {}:\n{}".format(
```
chekunkov: hm, does click.echo write in a wrong place?

vshlapakov: Yes, and sometimes it leads to bar duplication, with `click.echo`:
```
layers:   0%|                                                                                                                                            | 0/9Error The command '/bin/sh -c abcdef' returned a non-zero code: 127:
{'code': 127, 'message': "The command '/bin/sh -c abcdef' returned a non-zero code: 127"}
layers: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 9/9
Error: Build image operation failed
```
With `tqdm.write`:
```
Error The command '/bin/sh -c abcdef' returned a non-zero code: 127:
{'code': 127, 'message': "The command '/bin/sh -c abcdef' returned a non-zero code: 127"}
layers: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 9/9
Error: Build image operation failed
```
[tqdm.write](https://github.com/tqdm/tqdm#writing-messages) respects existing tqdm progress bars and updates it properly.

chekunkov: I've experienced a couple of similar issues with tqdm.write while working on push progress, maybe because there were multiple bars, but for the best result I'd recommend to close all the bars first before writing anything else - in this case click.echo should work properly as well.

vshlapakov: True, but this `click.echo` works inside an event loop, and there can be multiple non-fatal error events before operation stops and we're able to close the bar properly.

chekunkov: ah, I overlooked that, you're right

vshlapakov:
@jdemaeyer @chekunkov  I think I covered all the issues, any objections to squash and merge?

shub/image/build.py
```diff
@@ -71,6 +80,35 @@ def build_cmd(target, version, skip_tests):
...
+        # XXX: click.get_text_stream or click.get_binary_stream don't
+        # work well with tqdm on Windows and Python 3
+        file=sys.stdout,
+        # helps to update bars on resizing terminal
+        dynamic_ncols=True,
+        # miniters improves progress on erratic updates caused by network
+        miniters=1,
+        # don't need rate here, let's simplify the bar
+        bar_format='{l_bar}{bar}| {n_fmt}/{total_fmt}',
+    )
```
chekunkov: you can reuse https://github.com/scrapinghub/shub/blob/master/shub/image/push.py#L183 btw, say move to utils or something

vshlapakov:
@chekunkov I refactored the logic into classes like you proposed, hope it improved the changes.

jdemaeyer:
Code LGTM!

I don't like that we're always ignoring codecov. Though I can't think of any useful tests for this PR from the top of my head, do you have an idea?

shub/image/build.py
```diff
@@ -26,6 +26,7 @@
 """
 
 BUILD_STEP_REGEX = re.compile(r'Step (\d+)/(\d+) :*')
+BUILT_REGEX = re.compile(r'Successfully built ([0-9a-f]+)')
```
chekunkov: (nitpicking) maybe `BUILD_SUCCESS_REGEX`?

chekunkov:
@jdemaeyer yes, I was thinking how to test these changes and nothing useful came to my head. Best we can do is to mock docker client, return predefined logs, collect command output, compare to the reference output. but I'm not sure it'll guarantee the correctness of output in reality, I've seen weird issues with it on Py 2 when on Py 3 everything was fine, on Windows, when on Linux everything was correct. 🤷‍♂️

shub/image/build.py
jdemaeyer: While we're nitpicking anyways, I think the regex above this one is missing a `.` before the asterisk at the end. Unless you really wanted to match "an arbitrary number of colons" ;)

chekunkov: Good catch!

shub/image/build.py
```diff
@@ -49,19 +53,71 @@ def build_cmd(target, version, skip_tests):
...
+    Output all the events received from the docker daemon.
+    """
+    def handle_event(self, event):
+        if 'stream' in event:
+            self.handle_stream_event(event)
+        if 'error' in event:
+            tqdm.write("Error {}: {}".format(event['error'],
+                                             event['errorDetail']))
+            raise shub_exceptions.RemoteErrorException(
+                "Build image operation failed")
```
chekunkov: @vshlapakov is it a deliberate change? previously errors were not halting the progress https://github.com/scrapinghub/shub/blob/master/shub/image/build.py#L60

vshlapakov: Nah, you're right, missed it

chekunkov: What could be an example of the error event that shouldn't halt the progress though?

shub/image/build.py
```diff
@@ -49,19 +53,71 @@ def build_cmd(target, version, skip_tests):
...
+        if not step_row:
+            return
+        step_id, total = [int(val) for val in step_row.groups()]
+        if not self.bar:
+            self.bar = utils.create_progress_bar(
+                total=total,
+                desc='layers',
+                # don't need rate here, let's simplify the bar
+                bar_format='{l_bar}{bar}| {n_fmt}/{total_fmt}',
+            )
```
chekunkov: why not create it in init with total=1, then update total to the max(bar.total, total). this will remove a coulpe of `if not self.bar` checks

shub/image/build.py
```diff
@@ -49,19 +53,71 @@ def build_cmd(target, version, skip_tests):
...
+        # ignore onbuild sub-steps
+        if step_id > self.bar.pos and self.bar.total == total:
```
chekunkov: `self.bar.pos` is a bar position relative to the first bar in the multiple bars output. you wan't to use `self.bar.n`

chekunkov:
@vshlapakov can you quickly check how complex would it be to cover bars with tests, like I described above?

shub/image/build.py
vshlapakov: I recall while I tested this workflow in the past, I've seen some error events going one after another describing same error with more details, so I followed the stream model - process all the messages until there's something serious that stops yielding it. However I can't reproduce it now, and not sure if it's still applicable.

It was [discussed](https://github.com/docker/docker-py/issues/449) in `docker-py` tracker - and most users just check for `error`/`errorDetail` fields in event - that's what docker-compose [does](https://github.com/docker/compose/blob/master/compose/progress_stream.py#L56) and that's what we do when building an image on deploy.

After this investigation, I think it's fine to consider the change above as a deliberate one 😄  Objections?

chekunkov: Sounds good to me

chekunkov: Cannot we generalize `handle_event` and move it to the base class then?

shub/image/build.py
```diff
@@ -49,19 +53,71 @@ def build_cmd(target, version, skip_tests):
...
+            self.is_built = True
+            return
+        step_row = BUILD_STEP_REGEX.match(event['stream'])
+        if not step_row:
+            return
+        step_id, total = [int(val) for val in step_row.groups()]
+        if not self.bar:
+            self.bar = utils.create_progress_bar(
+                total=total,
+                desc='layers',
```
chekunkov: Just noticed - shouldn't it be 'Steps'? I was a bit confused when I tried it locally.

chekunkov: Also we may need some echo before build, like "Building {image}"

1__alex_mbp____dev_workspace_scrapinghub_repo_testspiders__zsh_

vshlapakov: Good idea 👍

shub/image/push.py
```diff
@@ -155,7 +151,7 @@ def show(self):
     def _create_total_bar(self):
         return utils.create_progress_bar(
             total=1,
-            desc='total',
+            desc='Total',
```
chekunkov: Layers?

tests/image/test_push.py
```diff
@@ -42,30 +43,18 @@ def test_cli_with_progress(docker_client_mock, test_mock):
...
-        'Layers:   0%|          | 0/2\r          \r'
-        'Layers:   0%|          | 0/3\r          \r'
-        'Layers:   0%|          | 0/3\r          \r'
-        'Layers:   0%|          | 0/3\n\r'
-        'abc:   0%|          | 0.00/24.8K [?B/s]\x1b[A\r'
-        '          \r'
-        'Layers:   0%|          | 0/3\n\n\r'
-        'egh:   0%|          | 0.00/26.3K [?B/s]\x1b[A\x1b[A\r'
-        '          \r'
+    assert re.search(
```
chekunkov: why not simply `assert reference_output in result.output`?

vshlapakov: Because progress speed (is a part of bar) depends on time and it's always different

tests/image/test_build.py
```diff
@@ -27,6 +31,31 @@ def test_cli(docker_client_mock, project_dir, test_mock):
...
+        {"stream": "some internal actions"},
+        {"stream": "Step 2/3 : RUN cmd1"},
+        {"stream": "some other actions"},
+        {"stream": "Step 3/3 : RUN cmd2"},
+        {"stream": "Successfully built 12345"}
+    ]
+    runner = CliRunner()
+    result = runner.invoke(cli, ["dev"])
+    assert result.exit_code == 0
+    placeholder = '█' if _supports_unicode(sys.stdout) else '#'
```
chekunkov: hm, interesting, why it's not the same for push?

vshlapakov: Good question, I'm on it (have just fixed tests for python 2 and broken for python 3) 😄

tests/image/test_push.py
```diff
@@ -29,6 +32,59 @@ def test_cli_with_apikey_login(docker_client_mock, test_mock):
...
+        'Layers:  67%\|██████▋   \| 2/3\r'
+        'Layers: 100%\|██████████\| 3/3\n\r'
+        'abc: 100%\|██████████\| 24.8K/24.8K \[[?.0-9]*[KM]?B/s\]\n\r'
+        'egh: 100%\|██████████\| 26.3K/26.3K \[[?.0-9]*[KM]?B/s\]\n'
+        'The image registry/user/project:test pushed successfully\.\n'
+    )
+    if ascii:
+        to_replace = {'█': '#', '▎': '3', '▋': '6'}
+        for k in to_replace:
+            expected = expected.replace(k, to_replace[k])
```
chekunkov: I think you can make a small test utility for this

chekunkov: Also I'd move ascii check inside it - so you would wrap output with some test function and it'll do all the replacement in the background if necessary. If we add more tests that will reduce amount of boilerplate code.

vshlapakov: Good point, I will 👍

chekunkov:
LGTM @vshlapakov, can you squash it now and I'll merge if @jdemaeyer has no pushback

jdemaeyer:
No pushback, good work!

PR #252, opened 2017-04-06
Add "Create scrapinghub.yml" wizard to shub image build

jdemaeyer:
* Move `update_yaml_dict` from `shub.config` into `shub.utils`
* Move onboarding wizard from `shub.deploy` into `shub.utils`
* In `shub image init`, don't look for `scrapinghub.yml`, but for `scrapy.cfg`
* Extend the onboarding wizard so that it also asks for a Docker registry (if called from a `shub image` command)
* In `shub image build`, call the onboarding wizard if we can't find a `scrapinghub.yml`, or can't find a Docker registry in it

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 import contextlib
 import datetime
+import errno
 import json
 import os
 import subprocess
@@ -21,6 +22,7 @@ from six.moves.urllib.parse import urljoin
 import click
 import pip
 import requests
+import yaml
 
 try:
     from scrapinghub import HubstorageClient
@@ -601,3 +603,34 @@ def download_from_pypi(dest, pkg=None, reqfile=None, extra_args=None):
     with patch_sys_executable():
         pip.main([cmd, '-d', dest, '--no-deps'] + no_wheel + extra_args +
                  target)
+
+
+@contextlib.contextmanager
+def update_yaml_dict(conf_path=None):
+    """
+    Context manager for updating a YAML file. Key ordering and comments are not
+    preserved.
+    """
+    if not conf_path:
+        click.secho("Using update_yaml_dict without path is deprecated. Import"
+                    " GLOBAL_SCRAPINGHUB_YML_PATH from shub.config",
+                    fg='yellow')
+        from shub.config import GLOBAL_SCRAPINGHUB_YML_PATH
+        conf_path = GLOBAL_SCRAPINGHUB_YML_PATH
+    try:
+        with open(conf_path, 'r') as f:
+            conf = yaml.safe_load(f) or {}
+    except IOError as e:
+        if e.errno != errno.ENOENT:
+            raise
+        conf = {}
+    # Code inside context manager is executed after this yield
+    yield conf
+    # Avoid writing "key: {}"
+    for key in list(conf):
+        if conf[key] == {}:
+            del conf[key]
+    with open(conf_path, 'w') as f:
+        # Avoid writing "{}"
+        if conf:
+            yaml.dump(conf, f, default_flow_style=False)
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -665,38 +665,63 @@ def get_project_dir():
         "Dockerfile in this directory or any of the parent directories.")
 
 
-def create_scrapinghub_yml_wizard(conf, target='default'):
+def create_scrapinghub_yml_wizard(conf, target='default', image=False):
     """
     Ask user for project ID, ensure they have access to that project, and save
     it to given ``target`` in local ``scrapinghub.yml`` if desired.
     """
     closest_sh_yml = os.path.join(get_project_dir(), 'scrapinghub.yml')
-    # Get default endpoint and API key (meanwhile making sure the user is
-    # logged in)
-    endpoint, apikey = conf.get_endpoint(0), conf.get_apikey(0)
-    project = click.prompt("Target project ID", type=int)
-    if not has_project_access(project, endpoint, apikey):
-        raise InvalidAuthException(
-            "The account you logged in to has no access to project {}. Please "
-            "double-check the project ID and make sure you logged in to the "
-            "correct acount.".format(project),
-        )
-    conf.projects[target] = project
-    if click.confirm("Save as default", default=True):
+    project = None
+    if target not in conf.projects:
+        # Get default endpoint and API key (meanwhile making sure the user is
+        # logged in)
+        endpoint, apikey = conf.get_endpoint(0), conf.get_apikey(0)
+        project = click.prompt("Target project ID", type=int)
+        if not has_project_access(project, endpoint, apikey):
+            raise InvalidAuthException(
+                "The account you logged in to has no access to project {}. "
+                "Please double-check the project ID and make sure you logged "
+                "in to the correct acount.".format(project),
+            )
+        # XXX: Save {'id': project} once we normalize project config on loading
+        conf.projects[target] = project
+    if image and not conf.get_target_conf(target).image:
+        repository = click.prompt(
+            "Image repository (leave empty to use Scrapinghub's repository)",
+            default=True, show_default=False)
+        if target == 'default':
+            conf.images[target] = repository
+        else:
+            # XXX: Remove once we normalize project config on loading
+            if not isinstance(conf.projects[target], dict):
+                conf.projects[target] = {'id': conf.projects[target]}
+            conf.projects[target]['image'] = repository
+    # Image repositories are always saved
+    if image or click.confirm("Save as default", default=True):
         try:
-            with update_yaml_dict(closest_sh_yml) as conf_yml:
-                default_entry = {'default': project}
-                if 'projects' in conf_yml:
-                    conf_yml['projects'].update(default_entry)
+            # XXX: Don't save the global config to not leak any configuration
+            #      options from the global config file into the project-
+            #      specific one. Instead, load only the local settings, then
+            #      update these
+            # XXX: Runtime import to avoid circular dependency
+            from shub.config import load_shub_config
+            local_conf = load_shub_config(load_global=False, load_env=False)
+            if project:
+                local_conf.projects[target] = project
+            if image:
+                if target == 'default':
+                    local_conf.images[target] = repository
                 else:
-                    conf_yml['projects'] = default_entry
-        except Exception:
+                    # XXX: Remove once we normalize project config on loading
+                    if not isinstance(local_conf.projects[target], dict):
+                        local_conf.projects[target] = {
+                            'id': local_conf.projects[target]}
+                    local_conf.projects[target].update({'image': repository})
+            local_conf.save(closest_sh_yml)
+        except Exception as e:
             click.echo(
-                "There was an error while trying to write to scrapinghub.yml. "
-                "Could not save project {} as default.".format(project),
+                "There was an error while trying to write to scrapinghub.yml: "
+                "{}".format(e),
             )
         else:
-            click.echo(
-                "Project {} was set as default in scrapinghub.yml. You can "
-                "deploy to it via 'shub deploy' from now on.".format(project),
-            )
+            click.echo("Saved to %s." % closest_sh_yml)
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -652,15 +652,25 @@ def has_project_access(project, endpoint, apikey):
             raise RemoteErrorException(str(e))
 
 
+def get_project_dir():
+    """Get the path to the closest directory that contains either
+    ``scrapinghub.yml``. ``scrapy.cfg``, or ``Dockerfile`` (in this priority).
+    """
+    for filename in ['scrapinghub.yml', 'scrapy.cfg', 'Dockerfile']:
+        closest = closest_file(filename)
+        if closest:
+            return os.path.dirname(closest)
+    raise NotFoundException(
+        "Cannot find project: There is no scrapinghub.yml, scrapy.cfg, or "
+        "Dockerfile in this directory or any of the parent directories.")
+
+
 def create_scrapinghub_yml_wizard(conf, target='default'):
     """
     Ask user for project ID, ensure they have access to that project, and save
     it to given ``target`` in local ``scrapinghub.yml`` if desired.
     """
-    closest_scrapycfg = closest_file('scrapy.cfg')
-    sh_yml_dir = (os.path.dirname(closest_scrapycfg) if closest_scrapycfg
-                  else os.getcwd())
-    closest_sh_yml = os.path.join(sh_yml_dir, 'scrapinghub.yml')
+    closest_sh_yml = os.path.join(get_project_dir(), 'scrapinghub.yml')
     # Get default endpoint and API key (meanwhile making sure the user is
     # logged in)
     endpoint, apikey = conf.get_endpoint(0), conf.get_apikey(0)
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -24,6 +24,8 @@ import pip
 import requests
 import yaml
 
+from scrapinghub import Connection, APIError
+
 try:
     from scrapinghub import HubstorageClient
 except ImportError:
@@ -634,3 +636,57 @@ def update_yaml_dict(conf_path=None):
         # Avoid writing "{}"
         if conf:
             yaml.dump(conf, f, default_flow_style=False)
+
+
+def has_project_access(project, endpoint, apikey):
+    """Check whether an API key has access to a given project. May raise
+    InvalidAuthException if the API key is invalid (but not if it is valid but
+    lacks access to the project)"""
+    conn = Connection(apikey, url=endpoint)
+    try:
+        return project in conn.project_ids()
+    except APIError as e:
+        if 'Authentication failed' in str(e):
+            raise InvalidAuthException
+        else:
+            raise RemoteErrorException(str(e))
+
+
+def create_scrapinghub_yml_wizard(conf, target='default'):
+    """
+    Ask user for project ID, ensure they have access to that project, and save
+    it to given ``target`` in local ``scrapinghub.yml`` if desired.
+    """
+    closest_scrapycfg = closest_file('scrapy.cfg')
+    sh_yml_dir = (os.path.dirname(closest_scrapycfg) if closest_scrapycfg
+                  else os.getcwd())
+    closest_sh_yml = os.path.join(sh_yml_dir, 'scrapinghub.yml')
+    # Get default endpoint and API key (meanwhile making sure the user is
+    # logged in)
+    endpoint, apikey = conf.get_endpoint(0), conf.get_apikey(0)
+    project = click.prompt("Target project ID", type=int)
+    if not has_project_access(project, endpoint, apikey):
+        raise InvalidAuthException(
+            "The account you logged in to has no access to project {}. Please "
+            "double-check the project ID and make sure you logged in to the "
+            "correct acount.".format(project),
+        )
+    conf.projects[target] = project
+    if click.confirm("Save as default", default=True):
+        try:
+            with update_yaml_dict(closest_sh_yml) as conf_yml:
+                default_entry = {'default': project}
+                if 'projects' in conf_yml:
+                    conf_yml['projects'].update(default_entry)
+                else:
+                    conf_yml['projects'] = default_entry
+        except Exception:
+            click.echo(
+                "There was an error while trying to write to scrapinghub.yml. "
+                "Could not save project {} as default.".format(project),
+            )
+        else:
+            click.echo(
+                "Project {} was set as default in scrapinghub.yml. You can "
+                "deploy to it via 'shub deploy' from now on.".format(project),
+            )
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -635,7 +635,7 @@ def update_yaml_dict(conf_path=None):
     with open(conf_path, 'w') as f:
         # Avoid writing "{}"
         if conf:
-            yaml.dump(conf, f, default_flow_style=False)
+            yaml.safe_dump(conf, f, default_flow_style=False)
 
 
 def has_project_access(project, endpoint, apikey):
--- a/freeze/tests/run.py
+++ b/freeze/tests/run.py
@@ -48,7 +48,7 @@ def test_version():
 def test_deploy_without_project():
     stdout, stderr = shub('deploy').communicate()
     assert stdout == b''
-    assert b'Error: No Scrapy project found in this location.' in stderr
+    assert b'Cannot find project' in stderr
 
 
 def test_deploy_default_project(apipipe, scrapyproject):
--- a/shub/config.py
+++ b/shub/config.py
@@ -1,5 +1,4 @@
 from __future__ import absolute_import
-import contextlib
 import netrc
 import os
 import warnings
@@ -14,7 +13,8 @@ from shub.exceptions import (BadParameterException, BadConfigException,
                              NotFoundException, ShubDeprecationWarning,
                              print_warning)
 from shub.utils import (closest_file, get_scrapycfg_targets, get_sources,
-                        pwd_hg_version, pwd_git_version, pwd_version)
+                        pwd_hg_version, pwd_git_version, pwd_version,
+                        update_yaml_dict)
 
 
 SH_IMAGES_REGISTRY = 'images.scrapinghub.com'
@@ -403,7 +403,7 @@ def _migrate_to_global_scrapinghub_yml():
         netrc_key = None
     if netrc_key:
         conf.apikeys['default'] = netrc_key
-    conf.save()
+    conf.save(GLOBAL_SCRAPINGHUB_YML_PATH)
     default_conf = ShubConfig()
     migrated_data = any(getattr(conf, attr) != getattr(default_conf, attr)
                         for attr in ('projects', 'endpoints', 'apikeys',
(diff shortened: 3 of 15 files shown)


jdemaeyer:
@vshlapakov @chekunkov Review please :) The first two commits are just moves, the next three change `shub image` functionality (see first comment)

shub/utils.py
```diff
@@ -570,3 +573,87 @@ def download_from_pypi(dest, pkg=None, reqfile=None, extra_args=None):
...
+        click.secho("Using update_yaml_dict without path is deprecated. Import"
+                    " GLOBAL_SCRAPINGHUB_YML_PATH from shub.config",
+                    fg='yellow')
+        from shub.config import GLOBAL_SCRAPINGHUB_YML_PATH
+        conf_path = GLOBAL_SCRAPINGHUB_YML_PATH
+    try:
+        with open(conf_path, 'r') as f:
+            conf = yaml.safe_load(f) or {}
+    except IOError as e:
+        if e.errno != 2:
```
chekunkov: please use `errno.ENOENT` for comparison

shub/utils.py
```diff
@@ -570,3 +573,87 @@ def download_from_pypi(dest, pkg=None, reqfile=None, extra_args=None):
...
+            raise
+        conf = {}
+    # Code inside context manager is executed after this yield
+    yield conf
+    # Avoid writing "key: {}"
+    for key in list(conf.keys()):
```
chekunkov: `for key in list(conf):` is a bit more pythonic - iteration over dict always returns keys

chekunkov:
> In shub image init, don't look for scrapinghub.yml, but for scrapy.cfg

I wonder whether it makes sense to check for scrapinghub.yml before any shub command, not only `shub deploy` and `shub image build`. E.g. if it's called and there's not closes scrapinghub.yml - then we ask user to configure "default" project.

shub/utils.py
redapple: is `list()` needed?

chekunkov: @redapple yes, because conf is modified during iteration

jdemaeyer:
Thanks Alex!

> I wonder whether it makes sense to check for scrapinghub.yml before all shub commands, not only shub deploy and shub image build

Some commands are not bound to a specific project, most importantly those that inspect jobs (`shub log`) or projects (`shub fetch-eggs`). It would be annoying if you couldn't run these outside of projects, e.g. in a freshly opened terminal

For those that are usually called inside a project (`shub image init`) but currently don't require a `scrapinghub.yml`, I don't see a big advantage of creating it "prematurely"

shub/utils.py
```diff
@@ -570,3 +573,87 @@ def download_from_pypi(dest, pkg=None, reqfile=None, extra_args=None):
...
+                "The account you logged in to has no access to project {}. "
+                "Please double-check the project ID and make sure you logged "
+                "in to the correct acount.".format(project),
+            )
+        conf.projects[target] = project
+    if image and target not in conf.images:
+        repository = click.prompt(
+            "Image repository (leave empty to use Scrapinghub's repository)",
+            default=True, show_default=False)
+        conf.images[target] = repository
```
vshlapakov: Remark: I'm fine with the changes but would like to clarify - according to the code wizard adds `(target, repo)` pair to `images` section, but I recall we agreed to move away from `images` section step-by-step, why not add it as `image: repo` under the specific project section?

vshlapakov:
@jdemaeyer LGTM, thanks!

chekunkov:
I guess we need to merge https://github.com/scrapinghub/shub/pull/253 first, otherwise the following exception occurs to me

```
→ shub image upload
fatal: Not a git repository (or any of the parent directories): .git
fatal: Not a git repository (or any of the parent directories): .git
fatal: Not a git repository (or any of the parent directories): .git
fatal: Not a git repository (or any of the parent directories): .git
Target project ID: 23661
Image repository (leave empty to use Scrapinghub's repository):
Saved to scrapinghub.yml.
  File "/Users/alex/.virtualenvs/hackershelf/bin/shub", line 11, in <module>
    load_entry_point('shub', 'console_scripts', 'shub')()
  File "/Users/alex/.virtualenvs/hackershelf/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/alex/.virtualenvs/hackershelf/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/alex/.virtualenvs/hackershelf/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/alex/.virtualenvs/hackershelf/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/alex/.virtualenvs/hackershelf/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/alex/.virtualenvs/hackershelf/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/upload.py", line 38, in cli
    build.build_cmd(target, version, skip_tests)
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/build.py", line 54, in build_cmd
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/utils.py", line 106, in format_image_name
    parts = image_name.rsplit('/', 1)
AttributeError: 'bool' object has no attribute 'rsplit'
```

vshlapakov:
> otherwise the following exception occurs to me

Good catch, it's caused by default value = `True` when asking for a repository name. We can focus on #253 for sure, other option is to just replace the default value with:

```
'images.scrapinghub.com/project/{}'.format(project_id)
```

jdemaeyer:
Good catch Alex. I think that is an issue in `shub image upload` and not in the wizard. Since we want to use `image: true` as a shortcut to use Scrapinghub's repo and since it's more handy I think that is what the wizard should save to the config.

chekunkov:
@jdemaeyer could you please rebase this PR on top of master? I'm going to give it another round of review and local testing.

jdemaeyer:
@chekunkov thanks, done!

jdemaeyer:
Ugh, sorry, I'll have to fix the failing tests tomorrow

jdemaeyer:
Alright, that was just a small fix

chekunkov:
@jdemaeyer I just tried it on the non-Scrapy project without scrapinghub.yml and instead of showing a wizard it failed with:

```
→ shub deploy
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Error: No Scrapy project found in this location.
```

Also I tried it on a Scrapy project that I used to run in the custom image - I removed scrapinghub.yml and called shub deploy - wizard appeared by it didn't ask me whether I want to use custom image for this project and after I gave it a project id it deployed it as a regular project.

```
→ shub deploy
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Target project ID: 23661
Save as default [Y/n]: Y
Saved to scrapinghub.yml.
Packing version r11-default
Deploying to Scrapy Cloud project "23661"
{"version": "r11-default", "project": 23661, "spiders": 1, "status": "ok"}
Run your spiders at: https://app.scrapinghub.com/p/23661/
```

I think we want wizard to configure images for the target

jdemaeyer:
Thanks for reviewing Alex.

My bad, I had mistakenly thought that `shub deploy` would act like `shub image push && shub image deploy`, so I had assumed that users must run `shub image build` first (where the wizard would then pop up).

I have updated `shub deploy` so that it'll show the wizard depending on which files exists and whether the target exists in `scrapinghub.yml`:

shub onboarding wizard behavior

chekunkov:
@jdemaeyer sorry, I've found a couple more issues

1. If I perform `shub deploy` and I have something in ~/.scrapinghub.yml but no scrapinghub.yml in the project dir - part of the global config leaks to the local config

```
→ cat scrapinghub.yml
images:
  default: true
  kumo-dev-demo: images.scrapinghub.com/project/108106
projects:
  default: 108106
  kumo-dev-demo: demo/108106
```

kumo-dev-demo is from global file:

```
→ cat ~/.scrapinghub.yml
projects:
    kumo-dev-demo: demo/108106

images:
    kumo-dev-demo: images.scrapinghub.com/project/108106
```

Is it unwanted side effect or it's expected and hard to fix?


2. If I perform `shub deploy` with a target that is missing from the config file I'll get a confusing error (not expected by users who don't use Scrapy)

```
→ shub deploy kumo-dev
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Error: No Scrapy project found in this location.
```

The scrapinghub.yml is the same as above. So definitely error should be different. And maybe shub should propose to create an entry for the missing target - but probably requires a separate ticket / PR. WDYT?

2.1. Actually the same error with removed scrpainghub.yml (I'd expect wizard to appear, there's Dockerfile and no scrapy.cfg)

3. If I don't have scrpainghub.yml locally and perform `shub image upload kumo-dev` wizard fails

```
→ shub image upload kumo-dev
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Target project ID: 108106
Image repository (leave empty to use Scrapinghub's repository):
  File "/Users/alex/.virtualenvs/kumo-forwarder/bin/shub", line 11, in <module>
    load_entry_point('shub', 'console_scripts', 'shub')()
  File "/Users/alex/.virtualenvs/kumo-forwarder/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/alex/.virtualenvs/kumo-forwarder/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/alex/.virtualenvs/kumo-forwarder/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/alex/.virtualenvs/kumo-forwarder/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/alex/.virtualenvs/kumo-forwarder/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/alex/.virtualenvs/kumo-forwarder/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/upload.py", line 39, in cli
    async, skip_tests)
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/upload.py", line 44, in upload_cmd
    build.build_cmd(target, version, skip_tests)
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/build.py", line 51, in build_cmd
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/utils.py", line 693, in create_scrapinghub_yml_wizard
    conf.projects[target]['image'] = repository
TypeError: 'int' object does not support item assignment

```

problem has its root in [this line](https://github.com/scrapinghub/shub/pull/252/files#diff-1e43fbb121ffe27f46a4e3a1cbe536b0R685) - int project id assigned here, so you either need to check that `conf.projects[target]` is dict and covert it to dict if its not, or assign dict `{'id': project}` from the beginning.

jdemaeyer:
Thanks again for the extensive testing Alex, it's clearly needed...

> part of the global config leaks to the local config
> Is it unwanted side effect or it's expected and hard to fix?

yeah :/ It's what I tried to fix in bbb8642, which at least avoids API key leaking. I had imagined having projects/images configured in the global `~/.scrapinghub.yml` to be a rare edge case but on second thought it's probably not as rare, particularly among users who are going to run the wizard often. I'll think of a different solution, possibly involving two instances of `ShubConfig` inside the wizard.

> If I perform shub deploy with a target that is missing from the config file I'll get a confusing error

I agree that the error message is confusing. It is because we only invoke the wizard when the target is `default` (as you pointed out). After skipping the wizard, because the target is not configured, it has the default of `image = False` and `shub` goes to the normal `shub deploy` behaviour, i.e. tries to upload a Scrapy project.

> And maybe shub should propose to create an entry for the missing target - but probably requires a separate ticket / PR. WDYT?

I don't know why we chose to skip the wizard unless the target is `default`. Possibly to make sure that it doesn't run when a numeric project ID is supplied. I think it would be safe to also call the wizard with other targets when `target.isdigit() == False`. It shouldn't cause any problems with backwards compatibility, after all we're just replacing a fatal error message with a wizard. If you agree I'll add it in this PR.

> problem has its root in this line - int project id assigned here, so you either need to check that conf.projects[target] is dict and covert it to dict if it's not, or assign dict {'id': project} from the beginning.

Man, the whole "some projects are `int` and others are `dict`" thing is just one big clusterfuck :( We should refactor it, always normalizing to a dictionary internally but keeping `prod: 33333` as syntactic sugar.

To summarize, I'll

1. change the "save" part in the wizard so that there's no leaks from the global config when saving. I'll do this by loading a separate instance `local_conf = load_shub_config(load_global=False, load_env=False)` inside the wizard, then saving it and calling `conf.load_file(closest_sh_yml)` on the global configuration. Ugly, but should work,
2. invoke the wizard on any non-existing, non-numerical targets,
3. fix the "convert int project to dict before setting `image` option" issue, and
4. create a new ticket for always using dicts for projects internally

chekunkov:
> I think it would be safe to also call the wizard with other targets when `target.isdigit() == False`.

@jdemaeyer sounds good to me. But there's another option - keep wizard for 'default' target only and return a better error if project id is not numeric and not present in the config - something like `target "xxx" is not configured in scrapinghub.yml, see how to configure here https://...`. Did you consider this approach, any downsides of picking it? A small usability problem with wizard on each non-numeric target is that typos in target will cause wizard to start. Not a big issue, but should be taken into account.

One thing about `shub deploy` invocations with `target.isdigit() is True` - right now it's always a Scrapy deploy, because shub doesn't check with the config in this case, right? It'll also fails with "Error: No Scrapy project found in this location." if you only have Dockerfile locally. Can we improve this somehow, and use numerical project ids to deploy images as well? Not clear to me how to distinguish `shub deploy 1234` for image and Scrapy in this case without a config, so maybe doesn't worth a hassle, but maybe you have some bright idea?

BTW, found interesting behavior of `shub image` commands in this branch when numerical project id is given, it shouldn't call a wizard in this case (check how master does that).

1__shub_image_upload_108106__python3_6_

jdemaeyer:
> A small usability problem with wizard on each non-numeric target is that typos in target will cause wizard to start

That's a very good point. It may have been the reason why we set the wizard to only run on the `default` target. Let's go with just fixing the error message.

> One thing about shub deploy invocations with target.isdigit() is True - right now it's always a Scrapy deploy, because shub doesn't check with the config in this case, right?

Hm, no, that's a bug. If there is an alias configured for project 1234, then `shub deploy alias` and `shub deploy 1234` should always yield the same behavior. That's what the [`ShubConfig.get_project()` function](https://github.com/scrapinghub/shub/blob/master/shub/config.py#L236) is for (called from `get_target_conf()`). It works for `stack` (I added a debug print):
```
jakob@MosEisley ~/shub/playground % cat scrapinghub.yml
projects:
  blah:
    id: 43100
    stack: scrapy:1.3-py3

jakob@MosEisley ~/shub/playground % shub deploy blah
Packing version 1.0
STACK: scrapy:1.3-py3
Deploying to Scrapy Cloud project "43100"

jakob@MosEisley ~/shub/playground % shub deploy 43100
Packing version 1.0
STACK: scrapy:1.3-py3
Deploying to Scrapy Cloud project "43100"
```

Looks like we introduced it with #253. [This `if` statement](https://github.com/scrapinghub/shub/blob/master/shub/deploy.py#L78) should be removed. I'll make a separate PR.

> Not clear to me how to distinguish shub deploy 1234 for image and Scrapy in this case without a config, so maybe doesn't worth a hassle, but maybe you have some bright idea?

Hm we already [have the necessary logic](https://github.com/scrapinghub/shub/blob/becd57c8f217422b93dcc518be89f721d4bbd4b5/shub/deploy.py#L98-L108) in this PR. It should be easy to generalize for this use case too

jdemaeyer:
Alright, I have

* fixed `TypeError: 'int' object does not support item assignment`
* fixed leaking of global config to `scrapinghub.yml` in the onboarding wizard
* fixed the error message on `shub deploy non-existent-target`. It will now no longer print "Not inside a Scrapy project" but our [standard "could not find target"](https://github.com/scrapinghub/shub/blob/master/shub/config.py#L255-L257) message
* fixed `shub image upload PROJECTID` so it won't call the wizard anymore (but instead complain that there is no image repository configured, should we change this so that it will assume SH's repo?)

I decided to skip "no `scrapinghub.yml`, no `scrapy.cfg`, but `Dockerfile`, then `shub deploy PROJECT_ID` issue we mentioned above. It's not as easy as I thought because we have no way of moving configuration from `shub deploy` to `shub image upload` except through the configuration file. I think it's enough of an edge case that we can justify always assuming `shub deploy PROJECT_ID` with no `scrapinghub.yml` means you're trying to deploy a Scrapy project.

I've already run a bunch of scenarios on my machine, but could you run it through another round of testing @chekunkov?

chekunkov:
1. why "fixed shub image upload PROJECTID so it won't call the wizard anymore"? if there's not scrapinghub.yml - wound't it be better to show wizard? e.g. compare the following 2 commands

```
→ shub image upload 108106
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Error: Not inside a project: scrapinghub.yml not found.
```

```
→ shub image upload
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Target project ID: 108106
Image repository (leave empty to use Scrapinghub's repository):
Saved to xxx/scrapinghub.yml.
Building images.scrapinghub.com/project/108106:f092796-fix-docker-disconnects.
...
```

The same with `shub image upload kumo-dev`.

The only difference is that in the first command user already provided target id, other than that cases are similar but first command fails and second does not.

What if we run wizard regardless of the presence of scrapy.cfg and Dockerfile? I have the following idea now - shub deploy or shub image build commands usually trigger wizard. Inside those commands we check if deploy is configured for the target either in ~/.scrapinghub.yml or in project scrapinghub.yml. If target is not configured we check whether project scrapinghub.yml exists. If it exists - shub always fails with standard "could not find target" message. If it doesn't exist - wizard always starts.

Now, depending on how shub commands was called we can modify what wizard asks:

- `shub deploy` - ask project id, "want to deploy as custom image?", if answer to the latter is yes - ask for image repository. save as default target
- `shub deploy 108106` - ask "want to deploy as custom image?", if answer to the latter is yes - ask for image repository. save as default target with project id 108106
- `shub deploy prod` - ask project id, "want to deploy as custom image?", if answer to the latter is yes - ask for image repository. save as 'prod' target with project id 108106. if later user will try to call `shub deploy` - shub will fail with "could not find target 'default'"

Note the "want to deploy as custom image?" question - it's a new question that we can add to remove guessing from the setup process. Should probably be reworded and contain link to custom image article in the help center. E.g. "Would you like to deploy your project as custom image? Read about that at http://..." For 'shub image *' calls we can skip this question as the answer will obviously be "yes" in that case. 

So far it seems intuitive enough for me, do you agree?

2. "I think it's enough of an edge case that we can justify always assuming shub deploy PROJECT_ID with no scrapinghub.yml means you're trying to deploy a Scrapy project."

Well, not really an edge case, because that's the thing we want to simplify for users who don't use Scrapy. Ideally users who want to deploy CasperJS crawler to the platform for the first time should be able to run either `shub deploy` or `shub deploy PROJECT_ID` or `shub deploy prod` and succeed with the help of the wizard. This can be confusing for them:

```
→ shub deploy 108106
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Packing version f092796-fix-docker-disconnects
Error: No Scrapy project found in this location.
```

or less confusing but definitely we can show wizard here just like we do for `shub image *` commands and save target under given name

```
→ shub deploy prod
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Usage: shub deploy [OPTIONS] [TARGET]

Error: Invalid value for target: Could not find target "prod". Please define it in your scrapinghub.yml or supply a numerical project ID.
```

3. Minor issue, but when shub decided to deploy Scrapy project, it asks whether scrapinghub.yml should be saved, but it always saves for image deploys. Wouldn't it be better to be consistent and either ask or don't ask in both cases?

I hope you are not tired of my comments yet :) We are close to finish this PR, other issues seem to be fixed.

jdemaeyer:
> why "fixed shub image upload PROJECTID so it won't call the wizard anymore"?

I was merely implementing your suggestion "it shouldn't call a wizard in this case (check how master does that)." ;)
For me it would be okay to run the wizard for `shub image upload 108106` if there is no `scrapinghub.yml` but a `Dockerfile`. But, why don't we just deploy the custom image to project 108106 without calling a wizard (using SH's repo)? That would match the behavior of `shub deploy 108106` on Scrapy projects.

> What if we run wizard regardless of the presence of scrapy.cfg and Dockerfile?

I don't necessarily like this because it would trigger the wizard if you run in the wrong directory -- just like we decided against `shub deploy unknown-target` triggering the wizard because it would get called everytime there is a typo. Besides, anything that can be deployed must have either `scrapy.cfg` or `Dockerfile`, what's the advantage of running a deploy wizard when there's no valid project to deploy?

> `shub deploy` - ask project id, "want to deploy as custom image?", if answer to the latter is yes - ask for image repository. save as default target

If I understand you correctly than that is pretty much as it is already implemented with this PR, [including the "want to deploy as custom image" question](https://github.com/scrapinghub/shub/blob/becd57c8f217422b93dcc518be89f721d4bbd4b5/shub/deploy.py#L101-L104).

> `shub deploy prod` - ask project id, "want to deploy as custom image?", if answer to the latter is yes - ask for image repository. save as 'prod' target with project id 108106. if later user will try to call shub deploy - shub will fail with "could not find target 'default'"

I thought we had decided against running the wizard on a given but unknown target so that it wouldn't run on typos?

> So far it seems intuitive enough for me, do you agree?

I don't think asking users about a "custom image" in the absence of a `Dockerfile` is intuitive. We need to remember that custom images are a power user feature, used by maybe ~5% of all customers (wild guess, actually we should get statistics on this). The confusion among beginner-level users who just want to deploy their Scrapy project (and probably never heard of Docker) and will then have to look up what an image is in the context of Scrapinghub is much bigger than the confusion among power users who know they're trying to deploy a custom Docker image to our platform and who see an unexpected "No scrapy project" when they run `shub deploy PROJECTID` instead of the `shub deploy` that was recommended to them through the docs.

> shub deploy 108106 - ask "want to deploy as custom image?", if answer to the latter is yes - ask for image repository. save as default target with project id 108106

> Well, not really an edge case, because that's the thing we want to simplify for users who don't use Scrapy.

As above, at no point in the docs should we ask them to run `shub deploy 108106` on a new project, but only ever for `shub deploy`. That being said, I agree that -- iff there is a `Dockerfile` and no `scrapy.cfg` -- `shub deploy 108106` should be an alias for `shub image upload 108106`.

> Minor issue, but when shub decided to deploy Scrapy project, it asks whether scrapinghub.yml should be saved, but it always saves for image deploys. Wouldn't it be better to be consistent and either ask or don't ask in both cases?

Yes. Ideally I would like to get rid of the "Save as default?" question and simply always save. After all, we're never overwriting any config through the wizard but only ever adding configuration.

This PR has become quite big by now. It's hard to review, and most of the code changes we seem to agree upon. I suggest we

1. Solve any discussions that would block merging the code as it is now,
2. If you agree that the current behavior is (maybe not optimal in your eyes, but) passable, clean up the commit history and merge this PR,
3. Create (a) new PR(s) to
    * Drop the "Save as default" question from the wizard
    * Deal with `shub deploy 108106` for custom image projects, making it an alias to `shub image upload 108106`
    * Deal with whether `shub deploy unknown-target` should call the wizard

jdemaeyer:
Ah, I forgot the most important part

> I hope you are not tired of my comments yet :) We are close to finish this PR, other issues seem to be fixed.

Not at all, I think the discussions are really fruitful =)

chekunkov:
> I was merely implementing your suggestion

good catch :) ok, let's focus on my latest comment for now - with the "What if we run wizard regardless of the presence of scrapy.cfg and Dockerfile" suggestions and your comments

> But, why don't we just deploy the custom image to project 108106 without calling a wizard (using SH's repo)? That would match the behavior of shub deploy 108106 on Scrapy projects.

I didn't know that shub deploy PROJECT_ID for Scrapy projects doesn't call wizard. But why don't call wizard in this case? no scrapinghub.yml, we have PROJECT_ID from command line, we can create scrapinghub.yml right there. any reason no to do that?

> I don't necessarily like this because it would trigger the wizard if you run in the wrong directory -- just like we decided against shub deploy unknown-target triggering the wizard because it would get called everytime there is a typo.

True, but I from my experience running shub deploy command from wrong directory is not that frequent. Don't think it'll cause much inconvenience, but that's my opinion of course. Typos in target were more frequent.

> Besides, anything that can be deployed must have either scrapy.cfg or Dockerfile, what's the advantage of running a deploy wizard when there's no valid project to deploy?

Ok, you're right here. We should check for scrapy.cfg or Dockerfile before wizard is triggered.

> If I understand you correctly than that is pretty much as it is already implemented with this PR

I see, it's called conditionally, missed that.

> I thought we had decided against running the wizard on a given but unknown target so that it wouldn't run on typos?

Here I was talking about the case when there's no scrapinghub.yml. Once you have a scrapinghub.yml - this will cause shub to return an error, see "If it exists - shub always fails with standard "could not find target" message. If it doesn't exist - wizard always starts." above. Try `shub image upload kumo-dev` in this PR, if there's not scrapinghub.yml it'll start wizard and save config with 'kumo-dev' target. `shub deploy` == `shub deploy default` - runs wizard and saves target default, so it seems consistent to me.

> I don't think asking users about a "custom image" in the absence of a Dockerfile is intuitive.

Already said above, I'm convinced, we should check for scrapy.cfg or Dockerfile before wizard is triggered

> We need to remember that custom images are a power user feature, used by maybe ~5% of all customers

we are working on shub changes now because we want make non-Scrapy projects support better, we are going to release something soon that can increase number of non-Scrapy users + increase number of non-power users among them, I'll share more details in private.

> As above, at no point in the docs should we ask them to run shub deploy 108106 on a new project

True, we shouldn't ask to do that in the docs, but as a user I think that if we support deploys with numeric project ids than wizard should work for the first deploy without scrapinghub.yml even if it's called like that.

>  Ideally I would like to get rid of the "Save as default?" question and simply always save

Let's remove it then :)

> I suggest we...

Sure, if it's easier for you - let's proceed like you suggested

PR #250, opened 2017-04-04
Create setup.py on shub image init

vshlapakov:
It makes more sense to create `setup.py` if it doesn't exist on `init` step instead of `build`.

Review, please.

PR #253, opened 2017-04-07
Shub deploy improvements

vshlapakov:
The goal of the PR is to improve logic of `shub deploy` to provide the following things:
- [x] modify shub config logic to process `image` per project
- [x] mark `images` section as deprecated and prefer `image` setting if defined
- [x] make `images.scrapinghub.com` a default registry and use SH apikey for it as an auth
- [x] use `shub image upload` instead of `shub deploy` when there's `image` defined for the project

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,16 +4,15 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-
 backports.ssl-match-hostname==3.5.0.1  # via docker-py
 click==6.6
 docker-py==1.10.6
 docker-pycreds==0.2.1     # via docker-py
-hubstorage==0.23.1
 ipaddress==1.0.17         # via docker-py
-PyYAML==3.12
+pyyaml==3.12
 requests==2.10.0
 retrying==1.3.3
 scrapinghub==1.9.0
 six==1.10.0
+tqdm==4.11.2
 websocket-client==0.37.0  # via docker-py
--- a/shub/config.py
+++ b/shub/config.py
@@ -16,6 +16,8 @@ from shub.utils import (closest_file, get_scrapycfg_targets, get_sources,
                         pwd_hg_version, pwd_git_version, pwd_version)
 
 
+SH_IMAGES_PREFIX = 'images.scrapinghub.com/project/'
+SH_IMAGES_REGISTRY = SH_IMAGES_PREFIX + '{project}'
 GLOBAL_SCRAPINGHUB_YML_PATH = os.path.expanduser("~/.scrapinghub.yml")
 NETRC_PATH = os.path.expanduser('~/_netrc' if os.name == 'nt' else '~/.netrc')
 
@@ -269,10 +271,36 @@ class ShubConfig(object):
             apikey=apikey,
             stack=(self.stacks.get(proj['stack'], proj['stack'])
                    if 'stack' in proj else self.stacks.get('default')),
+            image=self._select_image_for_project(target, proj),
             requirements_file=self.requirements_file,
             version=self.get_version(),
             eggs=self.eggs,
         )
+
+    def _select_image_for_project(self, target, project):
+        """Helper to select image for a project (or its target).
+
+        The select logic is the following:
+        - image defined per project has highest priority,
+        - images section is marked as deprecated, but still in force:
+          if target is defined in images - use a corresponding image,
+        - if image is not defined for project or target, stacks should
+          be used instead
+
+        The function responds with a custom image name string
+        (or None meaning regular stack-based deploy).
+        """
+        if target in self.images and target != 'default':
+            warnings.warn(
+                "Target is found in deprecated `images` section, please "
+                "replace it with global `image` setting or define `image` "
+                "setting for the project.")
+        image = (project['image'] if 'image' in project
+                 else self.images.get(target))
+        # aliases to use internal scrapinghub registry as image storage
+        if image is True or image == 'scrapinghub':
+            image = SH_IMAGES_REGISTRY.format(project=project['id'])
+        return image or None
 
     def get_target(self, target, auth_required=True):
         """Return (project_id, endpoint, apikey) for given target."""
@@ -295,15 +323,12 @@ class ShubConfig(object):
 
     def get_image(self, target):
         """Return image for a given target."""
-        if target not in self.images:
-            raise NotFoundException("Could not find image for %s. Please "
-                                    "define it in your scrapinghub.yml."
-                                    "" % target)
-        return self.images[target]
+        return self.get_target_conf(target, auth_required=False).image
 
 
 Target = namedtuple('Target', ['project_id', 'endpoint', 'apikey', 'stack',
-                               'requirements_file', 'version', 'eggs'])
+                               'image', 'requirements_file', 'version',
+                               'eggs'])
 
 
 MIGRATION_BANNER = """
@@ -457,3 +482,13 @@ def get_version():
     """Load shub configuratoin and return version."""
     conf = load_shub_config()
     return conf.get_version()
+
+
+def list_targets_callback(ctx, param, value):
+    """Click option callback to get targets from config, print it and exit."""
+    if not value:
+        return
+    conf = load_shub_config()
+    for name in conf.projects:
+        click.echo(name)
+    ctx.exit()
--- a/shub/config.py
+++ b/shub/config.py
@@ -16,8 +16,8 @@ from shub.utils import (closest_file, get_scrapycfg_targets, get_sources,
                         pwd_hg_version, pwd_git_version, pwd_version)
 
 
-SH_IMAGES_PREFIX = 'images.scrapinghub.com/project/'
-SH_IMAGES_REGISTRY = SH_IMAGES_PREFIX + '{project}'
+SH_IMAGES_REGISTRY = 'images.scrapinghub.com'
+SH_IMAGES_REPOSITORY = SH_IMAGES_REGISTRY + '/project/{project}'
 GLOBAL_SCRAPINGHUB_YML_PATH = os.path.expanduser("~/.scrapinghub.yml")
 NETRC_PATH = os.path.expanduser('~/_netrc' if os.name == 'nt' else '~/.netrc')
 
@@ -80,6 +80,13 @@ class ShubConfig(object):
                 option_conf = getattr(self, option)
                 yaml_option_conf = yaml_cfg.get(option, {})
                 option_conf.update(yaml_option_conf)
+                if option == 'images' and yaml_option_conf:
+                    click.echo(
+                        "Images section is deprecated, please replace it with "
+                        "global `image` setting or define `image` setting for "
+                        "the project.",
+                        err=True
+                    )
                 if shortcut in yaml_cfg:
                     # We explicitly check yaml_option_conf and not option_conf.
                     # It is okay to set conflicting defaults if they are in
@@ -288,19 +295,13 @@ class ShubConfig(object):
           be used instead
 
         The function responds with a custom image name string
-        (or None meaning regular stack-based deploy).
+        (or None/False meaning regular stack-based deploy).
         """
-        if target in self.images and target != 'default':
-            warnings.warn(
-                "Target is found in deprecated `images` section, please "
-                "replace it with global `image` setting or define `image` "
-                "setting for the project.")
-        image = (project['image'] if 'image' in project
-                 else self.images.get(target))
+        image = project.get('image', self.images.get(target))
         # aliases to use internal scrapinghub registry as image storage
         if image is True or image == 'scrapinghub':
-            image = SH_IMAGES_REGISTRY.format(project=project['id'])
-        return image or None
+            image = SH_IMAGES_REPOSITORY.format(project=project['id'])
+        return image
 
     def get_target(self, target, auth_required=True):
         """Return (project_id, endpoint, apikey) for given target."""
@@ -323,7 +324,23 @@ class ShubConfig(object):
 
     def get_image(self, target):
         """Return image for a given target."""
-        return self.get_target_conf(target, auth_required=False).image
+        target_conf = self.get_target_conf(target, auth_required=False)
+        project, image = target_conf.project_id, target_conf.image
+        if image is None:
+            raise NotFoundException(
+                "Could not find image for project '{}'. Please define it "
+                "in your scrapinghub.yml.".format(target))
+        elif image is False:
+            raise BadParameterException(
+                "Using custom images is disabled for the project '{}'. "
+                "Please enable it in your scrapinghub.yml.".format(target))
+        default_image = SH_IMAGES_REPOSITORY.format(project=project)
+        if image.startswith(SH_IMAGES_REGISTRY) and image != default_image:
+            click.echo(
+                "Found wrong SH repository for project '{}': expected {}.\n  "
+                "Please use aliases `True` or `scrapinghub` to fix it in your "
+                "config.".format(target, default_image), err=True)
+        return image
 
 
 Target = namedtuple('Target', ['project_id', 'endpoint', 'apikey', 'stack',
--- a/shub/config.py
+++ b/shub/config.py
@@ -73,6 +73,8 @@ class ShubConfig(object):
 
     def load(self, stream):
         """Load Scrapinghub configuration from stream."""
+        # flag to mark if images/default was used in the config file
+        check_default_image_scope = False
         try:
             yaml_cfg = yaml.safe_load(stream)
             if not yaml_cfg:
@@ -89,6 +91,8 @@ class ShubConfig(object):
                         .format(CONFIG_DOCS_LINK),
                         category=ShubDeprecationWarning
                     )
+                    if 'default' in yaml_option_conf:
+                        check_default_image_scope = True
                 if shortcut in yaml_cfg:
                     # We explicitly check yaml_option_conf and not option_conf.
                     # It is okay to set conflicting defaults if they are in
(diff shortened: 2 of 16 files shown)


vshlapakov:
@jdemaeyer @chekunkov Guys, I have some doubts about proper implementation and need your advice. It was discussed (and I assume - agreed) that `shub deploy` should launch `shub image deploy` instead if there's a custom image configured for the project/target. 

My main concerns are about merging/showing optional parameters: simplest option would be just adding all optional `shub image deploy` parameters to `shub image` parameters, marking it somehow (with some prefix I think) and deciding which logic to call based on shub config. But there're 6 parameters to add (`username`/`password`/`email`/`apikey`/`insecure`/`async`), that's a lot and I wonder if there's a known better way.

It's probably achievable with [click groups](http://click.pocoo.org/5/advanced/#command-aliases) or with something like [this](https://github.com/pallets/click/issues/72#issuecomment-42404559), but not sure if it's worth efforts as it will complicate logic. Thoughts?

chekunkov:
> shub deploy should launch shub image deploy

@vshlapakov `shub image upload`

chekunkov:
`async` is useless IMO, I'd deprecate and drop it over time

regards `username/password/email/apikey/insecure` - don't we assume that if we use `shub deploy` all of this should be somehow configured in the configuration file, or maybe docker login should be done if user is not yet authenticated to the registry? I don't think we should move all deploy parameters from `shub image upload` to `shub deploy`

vshlapakov:
> async is useless IMO, I'd deprecate and drop it over time
Agree, let's do it

About other parameters - you're right in a sense that all these parameters are optional and used only to "tune" deploy, for example in a case of a custom registry you need to provide custom credentials, it's not possible to setup it via configuration file. 

We may agree (and state it in the docs) that `shub deploy` could work as a shortcut to `shub image deploy` only if client is using our registry, does it makes sense?

chekunkov:
Discussed in a call - it seems like a only reasonable solution. `shub deploy` will be a shortcut that will build image, push it to our private registry (authenticated by apikey) and deploy the image in a single command. If user wants to use other registry he'll need to use `shub image *` commands. @dangra please tell if you have any pushback.

dangra:
that sounds perfectly reasonable @chekunkov

jdemaeyer:
Agreed

dangra:
Can we deprecate and eventually drop support to use registry other than scrapinghub's registry?

I think that feature was a usable while we didn't offer an internal registry with proper authentication, but now that users can push directly into our registry and don't have to pay for private repositories in registries like dockerhub or quay.io, I don't think it is needed anymore.

vshlapakov:
@dangra Great, I had doubts that we can afford it, that should simplify a lot of things 👍

shub/config.py
```diff
@@ -455,3 +480,12 @@ def get_version():
     """Load shub configuratoin and return version."""
     conf = load_shub_config()
     return conf.get_version()
+
+
+def list_targets(ctx, param, value):
```
vshlapakov: `list_targets` is used by `shub` and `shub.image` comands, and depends only on configuration file; let's move it here to avoid different import issues.

shub/utils.py
```diff
@@ -39,6 +39,24 @@
 STDOUT_ENCODING = sys.stdout.encoding or FALLBACK_ENCODING
 LAST_N_LOGS = 30
 
+_SETUP_PY_TEMPLATE = """\
```
vshlapakov: Creating setup.py from template was moved to `shub.utils` for similar reasons - it's used by `shub` and `shub.image` commands and keeping it in `shub.build` leads to import errors when 2 modules try to import each other.

vshlapakov:
The PR is ready for initial review, but I have to fix some minor issues (like adding missing docstrings and checking if all the cases are covered with tests).

requirements.txt
```diff
@@ -4,16 +4,15 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-
 backports.ssl-match-hostname==3.5.0.1  # via docker-py
 click==6.6
 docker-py==1.10.6
 docker-pycreds==0.2.1     # via docker-py
-hubstorage==0.23.1
```
vshlapakov: I guess we don't need `hubstorage` here, right?

shub/config.py
```diff
@@ -267,11 +269,37 @@ def get_target_conf(self, target, auth_required=True):
...
 
+    def _select_image_for_project(self, target, project):
+        """Helper to select image for a project (or its target).
+
+        The select logic is the following:
+        - image defined per project has highest priority,
+        - images section is marked as deprecated, but still in force:
+          if target is defined in images - use a corresponding image,
+        - if image is not defined for project or target, default image
+          should be used instead (shortcut to images->default),
```
chekunkov: Slightly confused by this because if image is not configured stacks should be used

shub/deploy.py
```diff
@@ -56,34 +60,11 @@
...
-        click.echo(name)
-    ctx.exit()
-
-
 @click.command(help=HELP, short_help=SHORT_HELP)
 @click.argument("target", required=False, default="default")
 @click.option("-l", "--list-targets", help="list available targets",
               is_flag=True, is_eager=True, expose_value=False,
-              callback=list_targets)
+              callback=list_targets_callback)
```
chekunkov: we need better documentation for list_targets (from the feedback)

shub/config.py
vshlapakov: Your comment just helped me to find a bug, thanks 👍

shub/utils.py
vshlapakov: Moved this change into https://github.com/scrapinghub/shub/pull/250, it's more consistent to change it there and rebase this PR on top of ot

shub/deploy.py
vshlapakov: Maybe `List available project names defined in your config`?
I used a part of `target` [definition](https://shub.readthedocs.io/en/stable/deploying.html#deploying-projects) from existing shub docs.

chekunkov:
Please rebase on top of master, #250 is merged

requirements.txt
jdemaeyer: Did you generate `requirements.txt` via `pip-compile`? On my machine it also removes the SSL backports and `ipaddress`.

shub/config.py
```diff
@@ -269,11 +271,37 @@ def get_target_conf(self, target, auth_required=True):
...
+        The function responds with a custom image name string
+        (or None meaning regular stack-based deploy).
+        """
+        if target in self.images and target != 'default':
+            warnings.warn(
+                "Target is found in deprecated `images` section, please "
+                "replace it with global `image` setting or define `image` "
+                "setting for the project.")
+        image = (project['image'] if 'image' in project
+                 else self.images.get(target))
```
jdemaeyer:
```python
image = project.get('image', self.images.get(target))
```
:)

shub/deploy.py
```diff
@@ -80,6 +74,19 @@ def list_targets(ctx, param, value):
...
+    conf, image = load_shub_config(), None
+    if target in conf.projects:
+        image = conf.get_target_conf(target).image
+    if not image:
+        deploy_cmd(target, version, debug, egg, build_egg, verbose, keep_log)
+    elif image.startswith(SH_IMAGES_PREFIX):
+        upload_cmd(target, version)
+    else:
+        raise BadParameterException("Please use `shub image` commands to work"
+                                    " with custom Docker registries.")
```
chekunkov: maybe " with Docker registries other than Scrapinghub default registry"?

shub/config.py
```diff
@@ -269,11 +271,37 @@ def get_target_conf(self, target, auth_required=True):
...
+        - image defined per project has highest priority,
+        - images section is marked as deprecated, but still in force:
+          if target is defined in images - use a corresponding image,
+        - if image is not defined for project or target, stacks should
+          be used instead
+
+        The function responds with a custom image name string
+        (or None meaning regular stack-based deploy).
+        """
+        if target in self.images and target != 'default':
```
chekunkov: why self.images['default'] doesn't raise a warning? I have config with 

```
images:
  default: xxx
```

it still should be deprecated and converted to `image: XXX`, right?

shub/image/list.py
```diff
@@ -52,31 +51,17 @@ def list_cmd_full(target, silent, version):
     image = config.get_image(target)
     version = version or config.get_version()
     image_name = utils.format_image_name(image, version)
```
chekunkov: we need to fix this exception 

```
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/build.py", line 47, in build_cmd
  File "/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/image/utils.py", line 96, in format_image_name
    parts = image_name.rsplit('/', 1)
AttributeError: 'NoneType' object has no attribute 'rsplit'
```

Happens when you don't have image config and use `shub image upload`

shub/config.py
vshlapakov: I think for now it's not possible to distinguish between `image: xxx` and `images: default: xxx` when working with `ShubConfig`, it requires some additional changes. 

In a word, `ShubConfig` [loads](https://github.com/scrapinghub/shub/blob/master/shub/config.py#L90) `image: xxx` as `self.images["default"]`, and [saves it back](https://github.com/scrapinghub/shub/blob/master/shub/config.py#L162) in the same manner, so `image` works like a "syntactic sugar" on the config file level.

jdemaeyer: It is [always saved back as `image:` and not as `images: default:`](https://github.com/scrapinghub/shub/blob/master/shub/config.py#L169-L171) (unless there's other entries in the `images` dictionary).

But for loading, indeed, after `ShubConfig.load()` has processed the YAML, it is impossible to distinguish whether the config file contained `image:` or `images: default:`. We could move this deprecation warning to `ShubConfig.load()` though.

shub/image/list.py
vshlapakov: I reverted some changes, now `config.get_image` raises different exceptions whether image is not found or image is False for a target, let me know if you disagree. The `get_image` function is used only by `shub.image` commands, and in my mind it should fail if image is expected but missing; otherwise we could add another `get_image_or_fail` with these exceptions.

shub/config.py
vshlapakov: > It is always saved back as image: and not as images: default:

Correct, sorry, I expressed it incorrectly.

> We could move this deprecation warning to ShubConfig.load()

+1

shub/config.py
```diff
@@ -16,6 +16,8 @@
                         pwd_hg_version, pwd_git_version, pwd_version)
 
 
+SH_IMAGES_PREFIX = 'images.scrapinghub.com/project/'
+SH_IMAGES_REGISTRY = SH_IMAGES_PREFIX + '{project}'
```
chekunkov: btw, I'd also add some check for correctness of sh_images path before all of `shub image` commands (maybe even in config) - I accidentally put repository name with wrong project id (valid project I have access to, but different project id from what it should deploy to)  and it worked :/ 

1__alex_mbp____dev_workspace_scrapinghub_repo_testspiders__zsh_

jdemaeyer: I wonder if there is a use case where that could be expected/wanted behavior (e.g. I want to use the same image for two different projects, perhaps because they have different project settings but are otherwise identical)

requirements.txt
vshlapakov: I did, it depends on which python you use to compile the dependencies list. On python 3 it drops `ipaddress` and SSL backports packages as you say, but for python 2 it keeps it as-is.

shub/config.py
vshlapakov: That's a good idea to validate repository if `images.sh.com` is used, I had a couple of similar issues in the past. Don't think there's such use case that could be wanted behaviour - that looks like a hidden bug, and can lead to pretty tricky cases. Also if image is already in our registry, pushing the same image with other tag is almost instant operation, as it will reuse all the layers already existing in the registry.

vshlapakov: I addressed it with https://github.com/scrapinghub/shub/pull/253/commits/e67d7f1cdfe645893746671701318d8753d33016, let me know wdyt.

jdemaeyer: LGTM

requirements.txt
jdemaeyer: Ahh that makes sense, let's keep it like this then

chekunkov:
1__shub_image_upload__python_

@vshlapakov can you fix this? too much noise, it should appear only once

chekunkov:
Also I think warning should point to the configuration docs, where this deprecation will be described as well, or at least there will be an example of the proper configuration

shub/config.py
```diff
@@ -295,15 +324,28 @@ def get_apikey(self, target, required=True):
...
+        elif image is False:
+            raise BadParameterException(
+                "Using custom images is disabled for the project '{}'. "
+                "Please enable it in your scrapinghub.yml.".format(target))
+        default_image = SH_IMAGES_REPOSITORY.format(project=project)
+        if image.startswith(SH_IMAGES_REGISTRY) and image != default_image:
+            click.echo(
+                "Found wrong SH repository for project '{}': expected {}.\n  "
+                "Please use aliases `True` or `scrapinghub` to fix it in your "
+                "config.".format(target, default_image), err=True)
```
chekunkov: I think it should halt execution, otherwise inattentive user can easily miss it + we agreed that it's a misconfiguration

chekunkov:
I think we should reword this exception a bit because it can be confusing for people who never used Scrapy and started with custom images

1__alex_mbp____dev_workspace_scrapinghub_repo_devbox_apps_kumo-forwarder_spider__zsh_

I tried it on a local non-Scrapy project and deliberately removed images section. I think exception should say something like "Cannot deploy project - failed to find Scrapy project and custom images are not configured in scrapinghub.yml".

chekunkov:
I think it's a bug:

```
→ shub deploy kumo-dev
Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
Error: No Scrapy project found in this location.
```

`image: true` should work for all targets

```
→ cat scrapinghub.yml
projects:
  default: devbox/1111112
  kumo-dev: 108106

image: true
```

vshlapakov:
@chekunkov That's a bug, but not an easy one, let's discuss it before trying to fix it. Like stated in this [thread](https://github.com/scrapinghub/shub/pull/253#discussion_r110708925), the configurations below mean exactly the same and it's not possible to distinguish between them for now after parsing the config (and except for: second one raises a warning):
```
image: true
```
```
images:
  default: true
```
But these 2 cases have different semantics, using `image` has global effect and applicable for all projects, while using `default` target in `images` section is in force only for `default` project (at least it works so now), do you agree? 

I don't have anything against global effect, but worrying about existing configs: if there's an image in `images/default` - it will be considered as a default image for all other projects. I'm inclined to modify `shub.config` a little bit, so `images` dictionary has some reserved key to store global `image` if defined and don't mix it with `images/default` one. Other option is to fail if `images/default` is found in config on start. Wdyt?

chekunkov:
@vshlapakov yeah, I'd somehow distinguish `images` from `image` and per project `image`. @jdemaeyer your opinion?

shub/config.py
```diff
@@ -81,11 +82,12 @@ def load(self, stream):
                 yaml_option_conf = yaml_cfg.get(option, {})
                 option_conf.update(yaml_option_conf)
                 if option == 'images' and yaml_option_conf:
-                    click.echo(
+                    warnings.warn(
```
vshlapakov: Addressing https://github.com/scrapinghub/shub/pull/253#issuecomment-293571505 and https://github.com/scrapinghub/shub/pull/253#issuecomment-293572056, I replaced click print to stderr with a usual warning.

But the real reason behind this issue is that each `load_shub_config()` command opens and parses a config file from scratch (sometimes even a few configs - global and local). Generally the workflow is fine per command, but if we call `shub deploy` with custom images - it's converted to a chain of commands, each one parsing same config again and again. A way better approach would be to read it only once, and pass parsed config to other commands as an optional parameter and load config file only if the param is not passed, I'll prepare the changes it if you agree.

shub/config.py
```diff
@@ -81,11 +82,12 @@ def load(self, stream):
...
                         "Images section is deprecated, please replace it with "
                         "global `image` setting or define `image` setting for "
-                        "the project.",
-                        err=True
+                        "the project. Check for additional details in {}."
+                        .format(CONFIG_DOCS_LINK),
+                        ShubDeprecationWarning
```
chekunkov: can we do the same with "Endpoint "%s" is still using HTTP." warning?

shub/config.py
jdemaeyer: I'm not sure if I like this. It spams the argument list of all commands for an edge case, and the time saved is negligible:

```
jakob@MosEisley ~/shub/playground % python -m timeit -s "from shub.config import load_shub_config" "load_shub_config()" 
1000 loops, best of 3: 1.61 msec per loop
```

vshlapakov: Ok then, warnings should do the trick 👌

jdemaeyer:
(Technically what I proposed violates semantic versioning, but using custom-image-based and SH-stack-based stacks in the same version control branch and explicitly wanting the default SH stack for non-default targets should be so much of an edge case that I think we can turn a blind eye this time)

shub/config.py
chekunkov: @vshlapakov @jdemaeyer what about memoizing the config upon first successful call? This way no extra args are needed, config is parsed only once.

I dislike how warnings look like when one calls the command compared to how it looked before.

```
→ shub deploy
/Users/alex/Dev/workspace/scrapinghub/repo/shub/shub/config.py:90: ShubDeprecationWarning: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project. Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
  ShubDeprecationWarning
```

chekunkov: Another option is to monkeypatch https://docs.python.org/3/library/warnings.html#warnings.showwarning

vshlapakov:
> after this PR, shub deploy prod will try to use a custom image

Right, because `images/default` = `image` enables image-based deploy mode for all projects.

> to fail if and only if projects has keys that are not found in images

That could work. So summarising the conditions to fail:
- non-empty `images` section
- default target is defined (there or via `image`), 
- `projects` section has keys that are not found in images. 

Looks good on the first side, it should fail on ambiguous cases.

chekunkov:
PR looks good to me now @vshlapakov. Should we cleanup commits a bit?

PR #259, opened 2017-04-12
Push progress bars - improve handling of push events

chekunkov:
For events without total progress take current as total.
For events with current progress greater than total - update total for the bar.

Fixes this:

push_empty_layer

### DIFF
--- a/tests/image/test_push.py
+++ b/tests/image/test_push.py
@@ -1,10 +1,15 @@
 # -*- coding: utf-8 -*-
-import re
+import inspect
+from functools import wraps
+
 import mock
 import pytest
 from click.testing import CliRunner
+
 from shub import exceptions as shub_exceptions
 from shub.image.push import cli
+from shub.image.utils import ProgressBar
+
 from ..utils import format_expected_progress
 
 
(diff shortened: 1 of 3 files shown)


tests/image/test_push.py
```diff
@@ -52,25 +77,72 @@ def test_cli_with_progress(docker_client_mock, test_mock):
...
+        'Layers: 100%|██████████| 4/4\n\r'
+        'abc: 100%|██████████| 24.8K/24.8K [1.00MB/s]\n\n'
+        'The image registry/user/project:test pushed successfully.\n'
     )
-    matched = re.search(expected, result.output)
-    assert matched
-    # to make sure the test regexp below is correct
-    assert matched.group(0).startswith('Layers:  33%')
-    assert matched.group(0).endswith('pushed successfully.\n')
+    assert expected in result.output
```
vshlapakov: Just curious: why not exact match? To skip potential warning messages at start?

tests/image/test_push.py
```diff
@@ -15,6 +20,26 @@ def test_mock():
         yield m
 
 
+@pytest.fixture(autouse=True)
+def monkeypatch_bar_rate(monkeypatch):
```
vshlapakov: All the monkeypatch logic is only to drop regexp below, right? I wasn't sure if it's worth it - as it brings more internals of tqdm, but It's less error-prone indeed and will be helpful to add more tests 👍

chekunkov: When I started writing new test I realized that converting output to regex manually is too much, I want something simpler.

tests/image/test_push.py
chekunkov: Correct, I have HTTP warnings that broke the tests

PR #260, opened 2017-04-13
Improve shub image deploy state reporting

chekunkov:
Mainly because of this
1__alex_mbp____dev_workspace_scrapinghub_repo_devbox_apps_kumo-forwarder_spider__zsh_

### DIFF
--- a/shub/image/deploy.py
+++ b/shub/image/deploy.py
@@ -105,50 +105,44 @@ def deploy_cmd(target, version, username, password, email,
         deploy_progress_cls = _LoggedDeployProgress
     else:
         deploy_progress_cls = _DeployProgress
-    deploy_progress = deploy_progress_cls(status_url)
+    events = _iterate_with_requests(status_url)
+    deploy_progress = deploy_progress_cls(events)
     deploy_progress.show()
 
 
-class _BaseDeployProgress(object):
-
-    def __init__(self, status_url):
-        self.status_url = status_url
-
-    def show(self):
-        while True:
-            event = _check_status_url(self.status_url)
-            self.handle_event(event)
+def _iterate_with_requests(status_url):
+    previous_event = None
+    while True:
+        event = _check_status_url(status_url)
+        # no need to handle already seen events
+        if previous_event != event:
+            yield event
             if event['status'] not in SYNC_DEPLOY_WAIT_STATUSES:
                 break
-            time.sleep(SYNC_DEPLOY_REFRESH_TIMEOUT)
-
-    def handle_event(self, event):
-        raise NotImplemented('Must be implemented in subclasses')
+            previous_event = event
+        time.sleep(SYNC_DEPLOY_REFRESH_TIMEOUT)
 
 
-class _LoggedDeployProgress(_BaseDeployProgress):
+class _LoggedDeployProgress(utils.BaseProgress):
     """Visualize deploy progress in verbose mode.
 
     Output all the distinct events received from the service.
     """
-    def __init__(self, status_url):
-        super(_LoggedDeployProgress, self).__init__(status_url)
-        self.previous_event = None
+    def show(self):
         click.echo("Deploy results:")
+        super(_LoggedDeployProgress, self).show()
 
     def handle_event(self, event):
-        if event != self.previous_event:
-            click.echo("{}".format(event))
-            self.previous_event = event
+        click.echo("{}".format(event))
 
 
-class _DeployProgress(_BaseDeployProgress):
+class _DeployProgress(utils.BaseProgress):
     """Visualize deploy progress in non-verbose mode.
 
     Uses a progress bar to track total progress.
     """
-    def __init__(self, status_url):
-        super(_DeployProgress, self).__init__(status_url)
+    def __init__(self, events):
+        super(_DeployProgress, self).__init__(events)
         self.progress_bar = self._create_progress_bar()
         self.result_event = None
 
--- a/shub/image/deploy.py
+++ b/shub/image/deploy.py
@@ -163,7 +163,7 @@ class _DeployProgress(_BaseDeployProgress):
         # last event with non-waiting status contains successful result or
         # error result from the service with error details
         if self.result_event:
-            click.echo("Deploy results:{}".format(self.result_event))
+            click.echo("Deploy results:\n{}".format(self.result_event))
 
     def handle_event(self, event):
         if 'progress' in event and 'total' in event:
--- a/shub/image/deploy.py
+++ b/shub/image/deploy.py
@@ -105,12 +105,17 @@ def deploy_cmd(target, version, username, password, email,
         deploy_progress_cls = _LoggedDeployProgress
     else:
         deploy_progress_cls = _DeployProgress
-    events = _iterate_with_requests(status_url)
+    events = _convert_status_requests_to_events(status_url)
     deploy_progress = deploy_progress_cls(events)
     deploy_progress.show()
 
 
-def _iterate_with_requests(status_url):
+def _convert_status_requests_to_events(status_url):
+    """Convert a sequence of deploy status requests to an iterator.
+
+    Current logic yields only distinct events and stops execution
+    when event status is a final one.
+    """
     previous_event = None
     while True:
         event = _check_status_url(status_url)
@@ -124,10 +129,8 @@ def _iterate_with_requests(status_url):
 
 
 class _LoggedDeployProgress(utils.BaseProgress):
-    """Visualize deploy progress in verbose mode.
+    """Visualize deploy progress in verbose mode."""
 
-    Output all the distinct events received from the service.
-    """
     def show(self):
         click.echo("Deploy results:")
         super(_LoggedDeployProgress, self).show()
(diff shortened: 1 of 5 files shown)


shub/image/deploy.py
```diff
@@ -79,51 +79,41 @@ def deploy_cmd(target, version, username, password, email,
...
         timeout=300,
         allow_redirects=False
     )
-    try:
-        req.raise_for_status()
-    except requests.exceptions.HTTPError:
-        _handle_deploy_errors(req)
-
-    click.echo("Deploy task results: {}".format(req))
+    if req.status_code == 400:
```
vshlapakov: Wonder if we should handle HTTP 401s as Dash can return it as well. Not sure if it makes sense - only if there's some significant timeout so token is expired, but I have never seen such cases before.

shub/image/deploy.py
```diff
@@ -79,51 +79,41 @@ def deploy_cmd(target, version, username, password, email,
...
+        reason = req.json().get('non_field_errors')
+        if reason:
```
vshlapakov: Maybe simplify it with?
```
raise ShubException('\n'.join(reason) if reason else req.content)
```

shub/image/deploy.py
```diff
@@ -79,51 +79,41 @@ def deploy_cmd(target, version, username, password, email,
...
+    req.raise_for_status()
     status_url = req.headers['location']
-
     status_id = utils.store_status_url(
         status_url, limit=STORE_N_LAST_STATUS_URLS)
     click.echo(
         "You can check deploy results later with "
         "'shub image check --id {}'.".format(status_id))
-
+    if async:
```
vshlapakov: I recall it was agreed to deprecate async mode, let's drop a warning somewhere?

shub/image/deploy.py
chekunkov: To my best knowledge Dash catches 401 from releaser and wraps it in the exception that produces a 400 error.

shub/image/deploy.py
chekunkov: Moved to a separate PR https://github.com/scrapinghub/shub/pull/262

chekunkov:
This is worrisome, 2nd time in a row I have to restart build because push progress tests fail - and I cannot reproduce that error locally. @jdemaeyer do tests pass fine locally for you?

jdemaeyer:
@chekunkov I've ran the tests locally on 2.7 and 3.6 three times each, no failures. What was the error on the failed builds?

chekunkov:
@jdemaeyer check the build now, it succeeded after rebase but after I hit "Restart build" button one of them failed.

jdemaeyer:
I had to restart builds in other PRs too so the errors seem unrelated to this PR. Maybe we can update the test to not expect exactly such a specific output but do multiple `assert in`s

vshlapakov:
@jdemaeyer Ok, now it's ready for next round of review :slightly_smiling_face: I extended Alex's work to provide a similar progress bar for image-deploy operation in non-verbose mode, please review and let me know if I missed something.

shub/image/deploy.py
```diff
@@ -79,51 +80,105 @@ def deploy_cmd(target, version, username, password, email,
...
+        # error result from the service with error details
+        if self.result_event:
+            click.echo("Deploy results:{}".format(self.result_event))
+
+    def handle_event(self, event):
+        if 'progress' in event and 'total' in event:
+            self.progress_bar.total = event['total']
+            self.progress_bar.update(max(event['progress'] - self.progress_bar.n, 0))
+        elif event['status'] not in SYNC_DEPLOY_WAIT_STATUSES:
+            self.result_event = event
```
jdemaeyer: Is there an advantage to repeating the `not in SYNC_DEPLOY_WAIT_STATUSES` and setting `result_event` instead of simply always setting `last_event` and using that in `show()`?

tests/image/test_deploy.py
```diff
@@ -135,12 +141,18 @@ def test_progress_verbose_logic(list_mocked, mocked_get, mocked_post, monkeypatc
...
-                " 'total': 100, 'last_step': 'updating panel'}") in result.output
-        # test for result in the end of output
-        assert result.output.endswith("{'status': 'ok', 'project': 1111112,"
-                                      " 'version': 'test', 'spiders': 10}\n")
+        lines = result.output.split('\n')
+        # test that output ends with a newline symbol
+        assert lines[-1] == ''
+        # test that the command succeeded
+        assert _load_deploy_event_result(lines[-2]) == {
+            'status': 'ok', 'project': 1111112, 'version': 'test', 'spiders': 10,
```
jdemaeyer: You could use `DEPLOY_EVENTS_BASE_SAMPLE[-1]` instead of repeating the dict here (and below)

tests/image/test_deploy.py
```diff
@@ -81,6 +81,7 @@ def test_cli_insecure_registry(list_mocked, post_mocked, get_mocked):
 
 @pytest.fixture
 def mocked_post(monkeypatch):
+    """Helper fixture to mock requests.post used for initial deploy request."""
```
jdemaeyer: I still wouldn't know what the fixture does after reading this ;) Maybe `Mock requests.post to return a status 200 response with location header`?

tests/image/test_push.py
```diff
@@ -56,7 +33,7 @@ def test_cli_with_apikey_login(docker_client_mock, test_mock):
 
 
 @pytest.mark.usefixtures('project_dir', 'test_mock')
-def test_cli_with_progress(docker_client_mock):
+def test_cli_with_progress(docker_client_mock, monkeypatch_bar_rate):
```
jdemaeyer: (You can also move it into `usefixtures` here)

shub/image/deploy.py
vshlapakov: Good question, I'm thinking about cases when network is broken in the middle of the deploy process (or there's something wrong with releaser service itself) so last event will not reflect deploy results, I have doubts it's worth to print it anyway.

shub/image/deploy.py
```diff
@@ -79,51 +80,105 @@ def deploy_cmd(target, version, username, password, email,
...
+
+    def __init__(self, status_url):
+        self.status_url = status_url
+
+    def show(self):
+        while True:
+            event = _check_status_url(self.status_url)
+            self.handle_event(event)
+            if event['status'] not in SYNC_DEPLOY_WAIT_STATUSES:
+                break
```
vshlapakov: Agree, I'll move it into an iterator :+1:

vshlapakov:
@jdemaeyer Thanks for such a nice review :+1: I think I addressed all the issues, does it look better now?

jdemaeyer:
Hey Viktor, looks good, thanks!

I have one more thing, but it's not super important so if you're short on time we can merge as is. The "progress printing" tests (`test_progress_verbose_logic()`, `test_progress_bar_logic()`, `test_progress_bar_logic_incomplete()`) essentially only test the `DeployProgress` classes, yet they go through the `cli` logic and therefore they have to set up a lot of mocks to avoid all the side effects and make all the I/O parts work as expected.

It would be cleaner if we tested these classes directly. That way we can simply give them a list or an iterator of the events without any mocks. We can check the output as we are now if we [define a mini-command in the test](https://github.com/scrapinghub/shub/blob/master/tests/test_utils.py#L513-L516) or alternatively via the [`capsys` fixture](https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function).

(Technically it would be even nicer if the `show()` methods wouldn't print anything but were an iterator yielding lines or progress bar updates, so we could test them without any mocking and capturing, but that's for another time. Sorry, I recently rewatched [this awesome talk on The Clean Architecture by Brandon Rhodes](https://www.youtube.com/watch?v=DJtef410XaM) so now I tend to look everywhere for chances to separate I/O from logic. ;))

vshlapakov:
Agree with all the aforementioned things, in the end used the "mini-command" approach and satisfied with results, the tests became cleaner, thanks for the proposal. Considering other approaches:

- yielding logs from `show()` method is better indeed, but it would require patching tqdm logic to yield logs, it's not straightforward because of tqdm asynchronous nature
- I haven't found how to use `capsys` fixture properly with unicode symbols

Thanks for the talk link btw :+1:

tests/image/test_deploy.py
```diff
@@ -130,17 +118,25 @@ def _load_deploy_event_result(line):
...
+        progress.show()
+    return progress_cmd
 
-    result = CliRunner().invoke(cli, ["dev", "--version", "test", "--verbose"])
-    assert result.exit_code == 0
 
+def _yield_events_with_timeout(events, timeout):
+    for event in events:
+        yield event
+        time.sleep(timeout)
```
jdemaeyer: Out of curiosity: What does the timeout do? Can't we just use `DEPLOY_EVENTS_BASE_SAMPLE` as `events` directly?

vshlapakov: This timeout is directly correlated with `tqdm` output, if you remove it - it will consume events instantly and print only initial (0%) and final (100%) progress instead of all the steps, I believe it's worth to keep the timeout to be closer to real-life conditions.

jdemaeyer: Makes sense, thanks for the explanation!

jdemaeyer:
Thanks @chekunkov and @vshlapakov, good work :)

PR #261, opened 2017-04-13
Improve Configuration docs page

jdemaeyer:


docs/configuration.rst
```diff
@@ -4,37 +4,161 @@
...
+        id: 12345
+        stack: scrapy:1.3-py3
+      prod: 33333
+
+The following is a list of all available configuration options:
+
+================ ============================================ =================
+Option           Description                                  Scope
+================ ============================================ =================
+``requirements`` Path to the project's requirements file, and global only
```
chekunkov: For the sake of source readability it'd be nice to have an extra white space between columns. I've read the line above as a whole sentence first :)

docs/configuration.rst
```diff
@@ -4,37 +4,161 @@
...
+ 
+    stack: scrapy:1.3-py3
+
+Using a custom Docker image::
+
+    projects:
+      default: 12345
+      prod: 33333
+
+    image: true
```
chekunkov: Could you add an example with global `image: true` and per-project `image: false` - to show how to deploy stack if image is enabled globally?

docs/configuration.rst
```diff
@@ -4,37 +4,161 @@
...
+----------------------
+
+Custom requirements file and fixed version information::
+
+    project: 12345
+    requirements:
+      file: requirements_scrapinghub.txt
+    version: 0.9.9
+
+Custom Scrapy Cloud stack, requirements file and and additional private
```
stummjr: there are two `and`s in the sentence above :)

docs/configuration.rst
```diff
@@ -4,37 +4,161 @@
...
-Cloud. By default, ``shub`` will auto-detect your version control system and
-use its branch/commit ID as version.
-
-Typically, you will configure only your API keys and the version specifier in
-the global ``~/.scrapinghub.yml``, and keep project configuration in the local
-``project_dir/scrapinghub.yml``. However, sometimes it can be convenient to
-also configure projects in the global configuration file. This allows you to
-use the human-friendly project identifiers outside of the project directory.
+
+.. _`Scrapy Cloud stack`: http://help.scrapinghub.com/scrapy-cloud/scrapy-cloud-stacks
```
stummjr: It's worth pointing to the new Knowledge base, instead of the old one: https://helpdesk.scrapinghub.com/support/solutions/articles/22000200402-scrapy-cloud-stacks

jdemaeyer:
Thanks @chekunkov, @stummjr :) Updated the PR

docs/configuration.rst
```diff
@@ -4,59 +4,197 @@
...
+        id: 12345
+        stack: scrapy:1.3-py3
+      prod: 33333
+
+The following is a list of all available configuration options:
+
+================  ============================================  ===============
+Option            Description                                   Scope
+================  ============================================  ===============
+``requirements``  Path to the project's requirements file, and  global only
```
vshlapakov: Very minor thing but still: I wonder is there any specific order for the options? Could we sort it by scope (global only at first, global/project then) or at least use alphabetical order?

docs/configuration.rst
```diff
@@ -4,59 +4,197 @@
...
+
+    projects:
+      default: 12345
+      staging: 33333
+      prod:
+        id: 44444
+        image: false
+        stack: scrapy:1.3-py3-20170322
+
+    image: true
```
vshlapakov: It was probably discussed earlier, but what if there's a global `image` and global `stack`? I'm pretty sure `image` has higher priority, but it can be non-obvious for clients.

chekunkov: Good question, probably should throw some configuration error

chekunkov:
@jdemaeyer is there anything you want to work on in this PR? If not - let's merge?

I think we might need to address https://github.com/scrapinghub/shub/pull/261#discussion_r112627655. Right now if both `stack` and `image` are configured globally then image deploy is used - but this is an ambiguous case and I think that some configuration error would be better. Can be done in a separate PR.

jdemaeyer:
@chekunkov Agreed, I'll prepare a PR

docs/configuration.rst
jdemaeyer: Ah, sorry Viktor, I forgot about this comment, I didn't mean to ignore you. ;) I have ordered them by how often they are used (based on guessing). `requirements` is less complex than `stack`, which is less complex than `image`. `version` is used rarely as the default makes a lot of sense, and `apikey` is usually not configured by hand.

I think this ordering makes sense if new users are curious about what can be configured. On the other hand, if you're not a new user but want to use the table as a reference, it's harder to find the option you're looking for.

vshlapakov: Ok, makes sense :+1:

PR #263, opened 2017-04-21
Fix shub deploy PROJECT_ID for configured projects

jdemaeyer:


shub/deploy.py
```diff
@@ -75,7 +75,7 @@
 @click.option("-k", "--keep-log", help="keep the deploy log", is_flag=True)
 def cli(target, version, debug, egg, build_egg, verbose, keep_log):
     conf, image = load_shub_config(), None
-    if target in conf.projects:
+    if target != 'default':
```
chekunkov: not sure I follow why we have this check at all. shouldn't we try to get target config always, use image if configured, catch error and set `image = None` if not cannot find target?

```
(Pdb) conf.get_target_conf(108106)
Target(project_id=108106, endpoint='https://app.scrapinghub.com/api/', apikey='f872157bf3f2481a9c58b7dc799ff836', stack=None, image='images.scrapinghub.com/project/108106', requirements_file=None, version='f092796-fix-docker-disconnects', eggs=[])
(Pdb) conf.get_target_conf(108107)
Target(project_id=108107, endpoint='https://app.scrapinghub.com/api/', apikey='f872157bf3f2481a9c58b7dc799ff836', stack=None, image='images.scrapinghub.com/project/108107', requirements_file=None, version='f092796-fix-docker-disconnects', eggs=[])
(Pdb) conf.get_target_conf('asdf')
*** shub.exceptions.BadParameterException: Could not find target "asdf". Please define it in your scrapinghub.yml or supply a numerical project ID.
```

vshlapakov: I guess it's a piece of my recent changes: the reason is that there's a case when getting target config is not needed at all, i.e. when user wants to build a project egg only with `--build-egg` [flag](https://github.com/scrapinghub/shub/blob/master/shub/deploy.py#L95). The deploy module code is "branchy" now, probably we could transform this build-egg case to click eager option with callback?

chekunkov: If that's the case I'd say it's definitely would be better to move build-egg code outside of the build cmd and remove some extra if-else's

jdemaeyer:
(I'm waiting on #252 to fix/update this, please don't merge)

jdemaeyer:
Superseded by #267. Note that there is no `target != 'default'` check in that PR.

PR #267, opened 2017-04-28
Refactor onboarding wizard

jdemaeyer:
When there's no `scrapinghub.yml`, `shub deploy 108106` will now no longer try to deploy to 108106 as Scrapy project (or fail with "No Scrapy project found"), but:

* if there is a `scrapy.cfg` and no `Dockerfile`, save 108106 as default target and deploy as Scrapy project
* if there is a `Dockerfile` and no `scrapy.cfg`, save 108106 as default target, ask for image repository, and deploy as custom image
* ask whether to use custom image if there are both `Dockerfile` and `scrapy.cfg`

The wizard will never touch any configuration if there already is a `default` target defined in `scrapinghub.yml`.

`shub image upload 108106` behaves similarly, but will always deploy as custom image project.

Closes #263

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -664,30 +664,46 @@ def get_project_dir():
         "Dockerfile in this directory or any of the parent directories.")
 
 
-def create_scrapinghub_yml_wizard(conf, target='default', image=False):
-    """
-    Ask user for project ID, ensure they have access to that project, and save
-    it to given ``target`` in local ``scrapinghub.yml`` if desired.
-    """
-    closest_sh_yml = os.path.join(get_project_dir(), 'scrapinghub.yml')
-    project = None
-    if target not in conf.projects:
-        # Get default endpoint and API key (meanwhile making sure the user is
-        # logged in)
-        endpoint, apikey = conf.get_endpoint(0), conf.get_apikey(0)
+def _get_target_project(conf, target):
+    """Ask for project ID (or use given target if numerical) and confirm that
+    user is logged in and has project access"""
+    # Will raise MissingAuthException if user is not logged in
+    endpoint, apikey = conf.get_endpoint(0), conf.get_apikey(0)
+    if target.isdigit():
+        project = int(target)
+        target = 'default'
+    else:
         project = click.prompt("Target project ID", type=int)
-        if not has_project_access(project, endpoint, apikey):
-            raise InvalidAuthException(
-                "The account you logged in to has no access to project {}. "
-                "Please double-check the project ID and make sure you logged "
-                "in to the correct acount.".format(project),
-            )
+    if not has_project_access(project, endpoint, apikey):
+        raise InvalidAuthException(
+            "The account you logged in to has no access to project {}. "
+            "Please double-check the project ID and make sure you logged "
+            "in to the correct acount.".format(project),
+        )
+    return target, project
+
+
+def _detect_custom_image_project():
+    """Guess if the user may want to deploy a custom image based on the
+    existence of ``scrapy.cfg`` and ``Dockerfile``. If there are both, ask."""
+    project_dir = get_project_dir()
+    has_scrapy_cfg = os.path.exists(os.path.join(project_dir, 'scrapy.cfg'))
+    has_dockerfile = os.path.exists(os.path.join(project_dir, 'Dockerfile'))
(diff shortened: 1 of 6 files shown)


jdemaeyer:
There is a bug somewhere in here. I have a project that has the default target defined and still I get `Saved to scrapinghub.yml.` and a modified (although equivalent) sh.yml

chekunkov:
@jdemaeyer maybe run wizard only if there's no scrapinghub.yml?

chekunkov:
Found one issue

```
There was an error while trying to write to /...scrapinghub.yml: local variable 'repository' referenced before assignment
```

shub/utils.py
```diff
@@ -669,9 +669,20 @@ def create_scrapinghub_yml_wizard(conf, target='default', image=False):
     Ask user for project ID, ensure they have access to that project, and save
     it to given ``target`` in local ``scrapinghub.yml`` if desired.
     """
+    if target.isdigit():
```
vshlapakov: Moving all the wizard logic into a helper function is a good step and I really like it. However the helper keeps growing and becomes too "branchy" (12 IFs in one function is a lot), it's hard to read, probably we could refactor it a bit and split to a few parts, wdyt?

I can clearly split it to 4 steps:
- checking if using wizard is required at all
- getting project id as input parameter (if needed)
- getting image repository as input parameter (if needed)
- updating local config file

Also it's possible that project is `None` and image is `False` [here](https://github.com/scrapinghub/shub/blob/ad3c8a97612e02877ce86c014f73ffd26036d426/shub/utils.py#L730), isn't it? I wonder if we should stop wizard if there're no changes.

jdemaeyer:
Thanks for the feedback @vshlapakov and @chekunkov!

After I got some distance to this over the weekend, and after considering your feedback, I have now done some heavier refactoring. The wizard is now "fire-and-forget", i.e. we don't have to figure out whether to call the wizard inside the commands but we just call `create_scrapinghub_yml_wizard()`. The wizard will then figure out if it needs to ask any questions. In particular, it will only do something if one of these conditions is met:

1. There is no local scrapinghub.yml
2. The wizard is called from one of the `shub image` commands, and the given target is defined but has no image repository configured. This use case is for transitioning a previously stack-based project to a custom image-based one, i.e. when you already have a `scrapinghub.yml` but now run `shub image build` for the first time

If none of these apply, it will do nothing and return.

Sorry for flip-flopping with this PR, but I think after this refactor the wizard is a lot less head-spinning, and in particular allows much cleaner tests.

Regarding splitting the wizard helper: I agree that it is a little "branchy", but these branches are not nested but independent of each other. So splitting the wizard into a couple of smaller helpers like `_ask_for_project_id` would simply mean moving once-indented code from the wizard to an extra function where it is still once-indented, with some extra logic necessary to move variables around. I think it got a bit more readable with the refactor, but that could also be because I've stared at it all day. ;) If you insist we can move the blocks I've now separated through blank lines into separate functions, but I don't think it would improve readability much.

chekunkov:
It worth splitting it into separate functions - it'll improve readability even more (there's still room for that IMO) + in my experience such flat uber functions tend to have hidden bugs because parts of the code that are functionally independent unintentionally leak some state in one another, and this cannot happen when such parts are limited by its function scope.

jdemaeyer:
Alrighty, wizard is split

chekunkov:
Tried `shub deploy prod` w/o scrapinghub.yml

```
→ shub deploy prod
WARNING: Images section is deprecated, please replace it with global `image` setting or define `image` setting for the project.
  Check for additional details in https://shub.readthedocs.io/en/stable/configuration.html.
Target project ID: 108106
Saving project 108106 as default target. You can deploy to it via 'shub deploy' from now on
Image repository (leave empty to use Scrapinghub's repository):
```

Config saved correctly but message above is misleading.

Resulting config 

```
→ cat scrapinghub.yml
projects:
  prod:
    id: 108106
    image: true
```

jdemaeyer:
Thanks, fixed

chekunkov:
@vshlapakov can you test it locally and tell if it works well? I everything if fine let's merge this and I'll finally update https://github.com/scrapinghub/shub/pull/249

vshlapakov:
@jdemaeyer @chekunkov It works well, let's merge it. Nicely done :+1:

PR #270, opened 2017-05-04
Fail on ambiguous image configuration

jdemaeyer:
Fail with an "ambiguous configuration" error message and a link for these three ambiguous configurations:

1. Both global non-false `image` and global `stack` configured
2. Global `image` configured and target project has `stack` option but not `image: false`
3. Target project has both `stack` option and non-false `image` option

Also replaces a couple of `BadParameterException`s that were raised on bad configurations with `BadConfigException`s (/cc @vshlapakov)

### DIFF
--- a/shub/config.py
+++ b/shub/config.py
@@ -353,6 +353,12 @@ class ShubConfig(object):
             raise BadConfigException(
                 "Using custom images is disabled for the project '{}'. "
                 "Please enable it in your scrapinghub.yml.".format(target))
+        elif target_conf.stack:
+            raise BadConfigException(
+                "Ambiguous configuration: There is both a custom image and a "
+                "stack configured for project '{}'. Please see {} for "
+                "information on how to configure both custom image-based and "
+                "stack-based projects.".format(target, CONFIG_DOCS_LINK))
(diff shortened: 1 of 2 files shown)


shub/config.py
```diff
@@ -350,12 +350,18 @@ def get_image(self, target):
                 "Could not find image for project '{}'. Please define it "
                 "in your scrapinghub.yml.".format(target))
         elif image is False:
```
chekunkov: curious - can `image` be `None` or empty string?

jdemaeyer: `image` is `None` when you haven't configured image deploy, e.g. for
```
project: 123
```

`image` is `False` when the user explicitly disabled image deploy, e.g. for
```
project: 123
image: false
```
I think Viktor separated `None` and `False` to give a more helpful error message to the user on why their command failed.

jdemaeyer:
```
projects:
  default: 123
  stackbased:
    id: 456
    image: false
    stack: scrapy:1.3

image: true
```

is probably a more real world example of why the separation is necessary ;) `shub image upload stackbased` will tell the user "using custom image is disabled for target 'stackbased'" instead of "could not find image"

PR #272, opened 2017-05-09
Make progress bar output tests independent of formatting details

jdemaeyer:
Our tests for the progress bar output during custom image building/pushing sometimes randomly fail. This is because the output of carriage returns, newlines, and ANSI escape sequences is dependent on the machine that runs the test.

For example, [this test](https://ci.appveyor.com/project/scrapinghub/shub/build/master-554/job/5e7mabanrd3cppe4#L138) failed because the actual output was `\x1b[A\\n\\x1b[AThe image registry/user/project:test pushed successfully.` while the expected output was `\\n\\nThe image registry/user/project:test pushed successfully.`

This PR should make the tests consistent by stripping these characters before checking the expected output.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
 import sys
-import mock
+import re
 
+import mock
 from click.testing import CliRunner
 from tqdm._utils import _supports_unicode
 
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -69,3 +70,8 @@ def format_expected_progress(progress):
         for sym in to_replace:
             progress = progress.replace(sym, to_replace[sym])
     return progress
+
+
+def clean_progress_output(output):
+    """Return output cleaned from \r, \n, and ANSI escape sequences"""
+    return re.sub('\n|\r|(\x1b\[|\x9b)[^@-_]*[@-_]|\x1b[@-_]', '', output)
```
chekunkov: the regex is hard to follow. can you visually split it and add comments using https://docs.python.org/2/library/re.html#re.X ?

jdemaeyer: That was actually a lot of fun, thanks :D

jdemaeyer: uhh nevermind, I seem to have broken it

jdemaeyer: fixed

PR #274, opened 2017-05-11
Add 'list-stacks' command

jdemaeyer:
#### Example outputs

Print only major versions by default
```
$ shub list-stacks
Scrapy stacks:
  scrapy:1.3
  scrapy:1.3-py3
  scrapy:1.2
  scrapy:1.2-py3
  scrapy:1.1
  scrapy:1.1-test
  scrapy:1.1-py3
  scrapy:1.0

Portia stacks:
  portia:0.13
```

Print regular versions with the `-a/--all` flag
```
% shub list-stacks -a
Scrapy stacks:                                                       
  scrapy:1.3
  scrapy:1.3-py3
  scrapy:1.3-py3-20170421
  scrapy:1.3-py3-20170322
  scrapy:1.3-py3-20170215
  scrapy:1.3-py3-20170209
  scrapy:1.3-py3-20170208
  scrapy:1.3-py3-20170203
  scrapy:1.3-20170421
  scrapy:1.3-20170322
  scrapy:1.3-20170215
[...]
```

Print URLs for manual lookup if there is an error while accessing the GitHub API
```
$ shub list-stacks
Error: Error while retrieving the list of stacks from GitHub: API rate
limit exceeded for 86.153.115.4. (But here's the good news: Authenticated
requests get a higher rate limit. Check out the documentation for more
details.)

Please visit the following URLs to see the available stacks: 
  Scrapy: https://github.com/scrapinghub/scrapinghub-stack-scrapy/releases
  Portia: https://github.com/scrapinghub/scrapinghub-stack-portia/releases
```

### DIFF
--- a/shub/list_stacks.py
+++ b/shub/list_stacks.py
@@ -1,0 +1,84 @@
+from __future__ import absolute_import
+
+import re
+
+import click
+import requests
+
+from shub.exceptions import RemoteErrorException
+
+
+RELEASES_WEB_URL = "https://github.com/{repo}/releases"
+TAGS_API_URL = "https://api.github.com/repos/{repo}/tags"
+STACK_REPOSITORIES = [
+    # (stack type, prefix, repository)
+    ('Scrapy', 'scrapy', 'scrapinghub/scrapinghub-stack-scrapy'),
+    ('Portia', 'portia', 'scrapinghub/scrapinghub-stack-portia'),
+]
+
+HELP = """
+List the available stacks that you can run your project in. Use --all to
+include regular releases.
+
+\b
+See
+https://helpdesk.scrapinghub.com/support/solutions/articles/22000200402-scrapy-cloud-stacks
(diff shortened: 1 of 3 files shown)


jdemaeyer:
Some notes for discussion:

1. GitHub paginates the tag list. Currently, `shub list-stacks` makes five requests (three for the Scrapy stacks, two for the Portia stacks) on every call, and this number will grow as we make more releases. Another advantage of having a Dash endpoint that lists the available stacks is that this would be avoided. (The "filter regular versions" logic could remain in shub I think, but I have no strong opinion on that.)
2. The `--all` list is pretty long. We could consider filtering it with an additional argument, e.g. `shub list-stacks scrapy:1.3` would only print stacks that begin with `scrapy:1.3`. In my eyes it's a bit over the top as `grep` provides the same functionality without having to look into the shub help page.

PR #275, opened 2017-05-11
Describe SHUB_JOB_DATA env variable

vshlapakov:
Please review and let me know if it can be improve in some way. There're some fields I haven't include in the list because think they are internal, also probably we should provide more real-life examples (at least for custom ``meta`` items if we're going to support it in the future).

### DIFF
--- a/docs/custom-images-contract.rst
+++ b/docs/custom-images-contract.rst
@@ -126,7 +126,8 @@ If you specified some custom metadata with ``meta`` field when scheduling the jo
 
 .. warning::
 
-    There could be some other fields but it's for internal use only and not a part of the contract.
+    ``SHUB_JOB_DATA`` may contain other undocumented fields. They are for the platform's internal use
+     and are not part of the contract, i.e. they can appear or be removed anytime.
 
 
 SHUB_SETTINGS
--- a/docs/custom-images-contract.rst
+++ b/docs/custom-images-contract.rst
@@ -90,14 +90,44 @@ SHUB_JOB_DATA
 
 Job arguments, in JSON format.
 
+
 **Example**:
 
 .. code-block:: javascript
 
     {"key": "1111112/2/2", "project": 1111112, "version": "version1",
-    "spider": "spider-name", "spider_type": "auto", "tags": [],
-    "priority": 2, "scheduled_by": "user", "started_by": "admin",
+    "spider": "spider-name", "spider_type": "auto", "tags": ["tagA", "tagB"],
+    "priority": 2, "scheduled_by": "user", "started_by": "john",
     "pending_time": 1460374516193, "running_time": 1460374557448, ... }
+
+
+Some useful fields
+__________________
+
+============ ======================================================== =================================
+Field        Description                                              Example
+============ ======================================================== =================================
+key          Job key in format ``PROJECT_ID/SPIDER_ID/JOB_ID``        ``"1111112/2/2"``
+project      Integer project ID                                       ``1111112``
+spider       String spider name                                       ``"spider-name"``
+spider_args  Spider args dictionary                                   ``{"arg1":"val1"}``
+version      String project version used to run the job               ``"version1"``
+units        Amount of units used by the job                          ``1``
+priority     Job priority value                                       ``2``
+tags         List of string tags for the job                          ``["tagA", "tagB"]``
+state        Job current state name                                   ``"running"``
+pending_time UNIX timestamp when the job was added, in milliseconds   ``1460374516193``
+running_time UNIX timestamp when the job was started, in milliseconds ``1460374557448``
+auth         Job authentication string to access job data             ``""eyJ0e***.eyJhd***.9H5Oq***"``
+scheduled_by Username who scheduled the job                           ``"john"``
+============ ======================================================== =================================
+
+If you specified some custom metadata with ``meta`` field when scheduling the job, the data will also be in the dictionary.
+
+.. warning::
+
+    There could be some other fields but it's for internal use only and not a part of the contract.
+
 
 SHUB_SETTINGS
 ^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


docs/custom-images-contract.rst
```diff
@@ -95,10 +95,39 @@ Job arguments, in JSON format.
...
+Field        Description                                              Example
+============ ======================================================== =================================
+key          Job key in format ``PROJECT_ID/SPIDER_ID/JOB_ID``        ``"1111112/2/2"``
+project      Integer project ID                                       ``1111112``
+spider       String spider name                                       ``"spider-name"``
+spider_args  Spider args dictionary                                   ``{"arg1":"val1"}``
+version      String project version used to run the job               ``"version1"``
+units        Amount of units used by the job                          ``1``
+priority     Job priority value                                       ``2``
+tags         List of string tags for the job                          ``["tagA", "tagB"]``
```
dangra: is it: JSON list of string tags for the job?

docs/custom-images-contract.rst
```diff
@@ -95,10 +95,39 @@ Job arguments, in JSON format.
...
+state        Job current state name                                   ``"running"``
```
dangra: state will be always `running` within a job environment, but it is good to document it anyway.

docs/custom-images-contract.rst
```diff
@@ -95,10 +95,39 @@ Job arguments, in JSON format.
...
+__________________
+
+============ ======================================================== =================================
+Field        Description                                              Example
+============ ======================================================== =================================
+key          Job key in format ``PROJECT_ID/SPIDER_ID/JOB_ID``        ``"1111112/2/2"``
+project      Integer project ID                                       ``1111112``
+spider       String spider name                                       ``"spider-name"``
+spider_args  Spider args dictionary                                   ``{"arg1":"val1"}``
+version      String project version used to run the job               ``"version1"``
```
dangra: is it version sent by user at deploy time or the incremental numeric id set by deploy service?

docs/custom-images-contract.rst
```diff
@@ -95,10 +95,39 @@ Job arguments, in JSON format.
...
+Some useful fields
+__________________
+
+============ ======================================================== =================================
+Field        Description                                              Example
+============ ======================================================== =================================
+key          Job key in format ``PROJECT_ID/SPIDER_ID/JOB_ID``        ``"1111112/2/2"``
+project      Integer project ID                                       ``1111112``
+spider       String spider name                                       ``"spider-name"``
+spider_args  Spider args dictionary                                   ``{"arg1":"val1"}``
```
dangra: In recent discussion we have been talking about using `job_cmd` for passing arguments to custom images even if it is not a script.  spider_args still need to be documented I think.

docs/custom-images-contract.rst
vshlapakov: First one (sent by user), though `deploy_id` is also there, now I think it's worth to document it too cause we show it in the deploy section

docs/custom-images-contract.rst
```diff
@@ -95,10 +95,39 @@ Job arguments, in JSON format.
...
+version      String project version used to run the job               ``"version1"``
+units        Amount of units used by the job                          ``1``
+priority     Job priority value                                       ``2``
+tags         List of string tags for the job                          ``["tagA", "tagB"]``
+state        Job current state name                                   ``"running"``
+pending_time UNIX timestamp when the job was added, in milliseconds   ``1460374516193``
+running_time UNIX timestamp when the job was started, in milliseconds ``1460374557448``
+auth         Job authentication string to access job data             ``""eyJ0e***.eyJhd***.9H5Oq***"``
```
redapple: Does `auth` need to be documented here? Can users use it somehow?

docs/custom-images-contract.rst
```diff
@@ -95,10 +95,39 @@ Job arguments, in JSON format.
...
+scheduled_by Username who scheduled the job                           ``"john"``
+============ ======================================================== =================================
+
+If you specified some custom metadata with ``meta`` field when scheduling the job, the data will also be in the dictionary.
+
+.. warning::
+
+    There could be some other fields but it's for internal use only and not a part of the contract.
```
redapple: Suggestion:
> `SHUB_JOB_DATA` may contain other undocumented fields. They are for the platform's internal use and are not part of the contract, i.e. they can appear or be removed anytime.

docs/custom-images-contract.rst
vshlapakov: There some Hubstorage endpoints that allow authentication via job authentication token, for example to get job data (items/requests/logs) from outside, sometimes it can be useful.

redapple: My point was that it's probably not very actionable for many/most users. It may trigger more questions than not documenting it (validity, where to use it, etc.)

vshlapakov: Valid point, I don't mind to drop it for now, and maybe return later when it will be supported on a whole client level.

docs/custom-images-contract.rst
jdemaeyer: Is this true? Of course the whole env var is a JSON-encoded string, but -- at least judging from the example above -- once it's been parsed, `tags` is a regular list and not a string containing a JSON list.

jdemaeyer:
I've added clarification comment [here](https://github.com/scrapinghub/shub/pull/275#discussion_r116949821), otherwise LGTM. Any objections to merging @dangra @redapple?

redapple:
Looks good to me @jdemaeyer @vshlapakov

docs/custom-images-contract.rst
vshlapakov: @jdemaeyer You're right - I was husty with this change, decided that the example is not clear enough w/o the JSON word, thanks for pointing it out. `tags` is a regular list after decoding indeed, as well as other compound structures (fields `job_cmd` is decoded as a list, `spider_args` as a regular dictionary) are decoded correspondingly. Examples are correct, I'm just not sure about proper description and how to represent the examples to make its format obvious enough. If current examples representation is clear then let's just drop `JSON` word from descriptions, do you agree?

jdemaeyer: Agreed, let's just drop the "JSON" here. We already mention that the whole environment var needs to be JSON-parsed at the top.

dangra: @jdemaeyer +1 on this.

PR #277, opened 2017-05-22
Added: Commands "jobmeta" and "stats"

### LINKED ISSUE #276: [idea] Support retrieving job stats / metadata
Something like `shub stats <job-id>` and `shub metadata <job-id>`

starrify:
Resolves #276


jdemaeyer:
Hi Pengyu, thanks a lot for this PR. :)

I had refrained from merging this so far because I would like to unify all the job-related commands under a `shub job` umbrella, hosting these two new commands as well as `items`, `log`, and `requests`.

However, now I think we should just get this out there and then work on the `shub job` umbrella for the next major release. Code LGTM, are there any last-minute concerns that riped over the last months from you? ;)

PR #278, opened 2017-05-24
Add 'bootstrap' command

jdemaeyer:
Add `shub bootstrap` command to clone example projects from https://github.com/scrapinghub/custom-images-examples.

~~I'll prepare an additional PR to the custom images repo that adds the `bootstrap_projects.yml` file.~~ `custom-images-examples` PR: https://github.com/scrapinghub/custom-images-examples/pull/10

#### Example outputs

Listing available projects:
```
$ shub bootstrap -l
Available projects:
  anemone   Download blog articles using the Anemone Ruby gem
  casperjs  Download arbitrary URLs and parse links using CasperJS. The
            description is really long, we should probably shorten it. Good
            thing we have output wrapping when listing projects.
  go        Download and parse test data from jsontest.com using Go
```

Cloning a project:
```
$ shub bootstrap casperjs
Downloading custom image examples
Cloning project 'casperjs' into /home/jakob/shub/casperjs/casperjs
```

Help page:
```
$ shub bootstrap -h
Usage: shub bootstrap [OPTIONS] PROJECT [TARGET_DIR]

  Through custom images, Scrapinghub allows you to run crawlers written in
  any language you want. To get you started, we prepared a few examples
  projects in different programming languages and frameworks. You can find
  them in our custom images repository at:

      https://github.com/scrapinghub/custom-images-examples

  The 'shub bootstrap' command clones an example project to the current
  directory so that you can start hacking right away.

  Run

      shub bootstrap -l

  to get a list of all available example projects, then run

      shub bootstrap PROJECTNAME

  to clone it.

Options:
  -l, --list  list available projects
  -h, --help  Show this message and exit.
```


shub/bootstrap.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@click.command(help=HELP, short_help=SHORT_HELP)
+@click.option('-l', '--list', 'list_projects', help='list available projects',
+              is_flag=True, callback=list_projects_callback,
+              expose_value=False, is_eager=True)
+@click.argument('project')
+@click.argument('target_dir', required=False)
+def cli(project, target_dir):
+    target_dir = os.path.join(os.getcwd(), target_dir or project)
```
chekunkov: maybe use clone project to the current directory if target_dir is not specified? i.e. current directory becomes a root, as if target dir was empty string `""`. if target is specified we join it like above, although I would also wrap it in `os.path.normpath`

shub/bootstrap.py
```diff
@@ -0,0 +1,137 @@
...
+    return zipfile.ZipFile(BytesIO(resp.content))
+
+
+def unzip_project(repo_zip, project, target_dir):
+    filenames = repo_zip.namelist()
+    repo_dirname = filenames[0]
+    project_filenames = [
+        fn
+        for fn in filenames
+        if fn.startswith(repo_dirname + project['directory'])
```
chekunkov: `s/directory/path/` ?

jdemaeyer: absolutely

shub/bootstrap.py
jdemaeyer: I had thought about this and even implemented it at some point. In the end I decided for always cloning into a new subdir for two reasons:

* This is what `git clone` and `scrapy startproject` do
* It is less annoying to fix "it cloned into a subdir when I wanted the current dir" (by just moving everything one level up) than it is to fix "it cloned into the current dir but I wanted a subdir", because if you expect a subdir you will likely run the command in a generic folder where you keep multiple projects (e.g. I personally have `~/python/`), and now you have to figure out which files were there before and which were just cloned.

Although I am open to being convinced otherwise if you have a strong argument ;)

About the `normpath` I am curious, I've seen it in other projects too. Is there a case where `os.path.join(os.getcwd(), 'some_folder')` gives an invalid/unwanted path? From what I've gathered around the internets the "cleaning up redundant separators" part [can screw up paths that contain symlinks](https://stackoverflow.com/a/6394943).

jdemaeyer:
`custom-images-examples` PR: https://github.com/scrapinghub/custom-images-examples/pull/10

shub/bootstrap.py
chekunkov: 1. re subdir - sounds reasonable, I don't have stronger arguments
2. re normpath - I proposed it because it provides a better user experience if target dir path is relative and contains links to parent directories. if you specify target as `../../path` it's easier for user to read `/path` rather than `/parent/current/../../path`. IMO humans should not try to evaluate final path in their heads, this is error prone. if you're concerned about symlinks let's use `os.path.realpath` instead.

jdemaeyer: Makes sense. On second thought I opted for `normpath`

shub/bootstrap.py
```diff
@@ -0,0 +1,138 @@
...
+    import os.path
+    projects_path = os.path.join(
+        os.path.dirname(__file__), 'bootstrap_projects.yml')
+    with open(projects_path, 'r') as f:
+        return yaml.safe_load(f)
+
+
+if not os.getenv('USING_TOX'):
+    get_available_projects = _get_available_projects  # noqa
+# XXX: END Remove this
```
chekunkov: https://github.com/scrapinghub/custom-images-examples/pull/10 is merged, please cleanup the PR @jdemaeyer

jdemaeyer:
@chekunkov Removed the debug part and added a sentence to the `custom-images-contract` docs. Please merge if you think it's okay

Issue #269, opened 2017-04-28
Release new major version


jdemaeyer:
shub's codebase contains a lot of deprecated commands/functions, and code that only serves backwards compatibility. I would like to aggressively remove this code to tidy up the codebase and improve the maintenance experience.

This ticket is to collect and discuss ideas and requirements for this release.

Changes that impact behavior of `shub` as command line client:

* Remove all deprecated commands and flags
* Remove support for `images` and `stacks` sections in `scrapinghub.yml`

Changes that impact behavior of `shub` as library:

* Remove all deprecated functions
* Clean up `shub.utils`
* Aggressively clean up `ShubConfig`, this includes the removal of `.normalized_projects` and _all_ the `get_xyz` methods including `get_target_conf` and `get_target`. Optimally I would like the only public methods to eb the load/save methods and `__contains__()` and `__getitem__()` for configuration retrieval. The latter must support getting a configured project by ID and not by alias
* Unify the layout of commands. Currently, some commands have most of their logic in their `@click.command`, while others have most of their logic in a separate function (without `click` decorators)

Internal changes:

* The configuration should never be loaded more than once. We can use `click`s `Context` to ensure this without having to always pass it around
* Tests should use the same framework. Currently we have some `unittest`-based tests and some `pytest`-based tests. I would like to write all tests based on `pytest`

ariasuni:
Maybe this should be updated to the state of shub 2.8.0 with some checked/unchecked checkboxes?

vshlapakov:
Unfortunately there was no active development during last year, and 2.8.0 doesn't address any of the issues/features listed here - it's rather about maintaining, bugfixing and merging contributions. However converting this to a list of pending tasks with related issues/PRs is a good idea and we'll definitely do it in the future.

PR #279, opened 2017-05-25
Replace list-spiders with shub-image-info

vshlapakov:
We're introducing a new command to Kumo contract: it's called `shub-image-info` and will deprecate current `list-spiders`. Comparing with `list-spiders` command - the new command should return JSON encoded metadata for a given image, incl. a project type, spiders metadata and probably some other fields in the future. As now we're changing the way how arguments and settings are being passed to a spider/script, scripts become a Scrapy-specific feature, and should be included into the same spiders list with `py:` prefix.

Important note, that we must provide backward compatibility, so if `shub-image-info` is not found, there's fall back to `list-spiders` with a deprecation warning.

Please, review.

### DIFF
--- a/shub/image/list.py
+++ b/shub/image/list.py
@@ -74,14 +74,14 @@ def list_cmd(image_name, project, endpoint, apikey):
     elif exit_code == 127:
         # FIXME we should pass some value for SCRAPY_PROJECT_ID anyway
         # to handle `scrapy list` cmd properly via sh_scrapy entrypoint
-        environment['SCRAPY_PROJECT_ID'] = str(project) if project else ''
+        # environment['SCRAPY_PROJECT_ID'] = str(project) if project else ''
         exit_code, logs = _run_cmd_in_docker_container(
             image_name, 'list-spiders', environment)
         if exit_code != 0:
             click.echo(logs)
             raise ShubException('Container with list cmd exited with code %s' % exit_code)
         logs = logs.decode('utf-8') if isinstance(logs, binary_type) else logs
-        return {'spiders': utils.valid_spiders(logs)}
+        return {'spiders': utils.valid_spiders(logs.splitlines())}
     else:
         click.echo(logs)
         raise ShubException(
@@ -145,4 +145,7 @@ def _extract_metadata_from_image_info_output(output):
         if not (name and isinstance(name, string_types)):
             raise_shub_image_info_error("spider name can't be empty or non-string")
         scripts.append(name[3:]) if name.startswith('py:') else spiders.append(name)
-    return {'spiders': spiders, 'scripts': scripts}
+    return {
+        'spiders': utils.valid_spiders(spiders),
+        'scripts': utils.valid_spiders(scripts),
+    }
--- a/shub/image/list.py
+++ b/shub/image/list.py
@@ -1,11 +1,13 @@
 import json
 
 import click
+import docker
 import requests
 from six import binary_type
+from six import string_types
 from six.moves.urllib.parse import urljoin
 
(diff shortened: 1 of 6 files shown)


chekunkov:
Tested with https://github.com/scrapinghub/custom-images-examples/pull/9, LGTM in general

shub/image/list.py
```diff
@@ -52,27 +54,38 @@ def list_cmd_full(target, silent, version):
...
-    if status_code != 0:
+    environment = {'JOB_SETTINGS': json.dumps(settings)}
+    exit_code, logs = _run_cmd_in_docker_container(
+            image_name, 'shub-image-info', environment)
+    if exit_code == 0:
+        return _extract_metadata_from_image_info_output(logs)
+    # shub-image-info command not found, fallback to list-spiders
+    elif exit_code == 127:
+        # FIXME we should pass some value for SCRAPY_PROJECT_ID anyway
+        # to handle `scrapy list` cmd properly via sh_scrapy entrypoint
```
jdemaeyer: (I know this `FIXME` was just copied over, but) What happens when there is no/an empty `SCRAPY_PROJECT_ID` defined, as implemented?

shub/image/list.py
```diff
@@ -89,27 +102,47 @@ def _get_project_settings(project, endpoint, apikey):
...
+        raise_shub_image_info_error('output is not a valid JSON dict')
+    if not isinstance(spiders_list, list):
+        raise_shub_image_info_error('spiders section must be a list')
+
+    spiders, scripts = [], []
+    for name in spiders_list:
+        if not (name and isinstance(name, string_types)):
+            raise_shub_image_info_error("spider name can't be empty or non-string")
+        scripts.append(name[3:]) if name.startswith('py:') else spiders.append(name)
+    return {'spiders': spiders, 'scripts': scripts}
```
jdemaeyer: For `list-spiders` (and in the fallback to it in line 84 above), we ran the spider list through `utils.valid_spiders()`. Is there a reason to skip it here?

vshlapakov: No reason, it's a good idea to validate spiders/scripts lists with the same regex, thanks :+1:

shub/image/list.py
vshlapakov: Earlier `scrapinghub-entrypoint-scrapy` relied on `SCRAPY_PROJECT_ID` to decide between a spider and a python script, later we changed it with https://github.com/scrapinghub/scrapinghub-entrypoint-scrapy/pull/26 in version `0.8.8`, but there may be a couple of custom images in-use relying on previous sh-ep-scrapy versions. So if there's no `SCAPY_PROJECT_ID` for such cases, `scrapy list` will be launched incorrectly (as a package script) and fail with not-found-error or something like this.

Now I wonder how many projects use custom images with oldy `scrapinghub-entrypoint-scrapy` versions in real, maybe it's already fine to finally drop it. That's a shame that PyPi disabled download stats :confused:

shub/image/list.py
vshlapakov: Addressed with https://github.com/scrapinghub/shub/pull/279/commits/e8e0989f66fad911287728c13b9ba5458a7e66cc

shub/image/list.py
jdemaeyer: Hm I see. Let's wait for #269 to drop it

vshlapakov:
Thanks @jdemaeyer! :slightly_smiling_face: Additional note - I believe we should release `shub` with the changes before we could merge https://github.com/scrapinghub/custom-images-examples/issues/9, are there any ETA for next `2.6.2`? It's not urgent, just to get some estimation.

PR #280, opened 2017-05-25
Extend contract with shub-image-info

vshlapakov:
Describe extending custom images contract with a new ``shub-images-info`` command.

The PR is complete, awaiting for next shub release and update examples in custom-images-examples repo.

PR #286, opened 2017-06-15
Extract and pass project_type on deploy

vshlapakov:
shub tool should extract `project_type` info from `shub-image-info` and pass it on deploy.

Review, please.

### DIFF
--- a/shub/image/list.py
+++ b/shub/image/list.py
@@ -137,13 +137,15 @@ def _extract_metadata_from_image_info_output(output):
 
     try:
         metadata = json.loads(output)
-        spiders_list = metadata.get('spiders', [])
-    except (ValueError, AttributeError):
+        project_type = metadata.get('project_type')
+    except (AttributeError, ValueError):
         raise_shub_image_info_error('output is not a valid JSON dict')
-    if not isinstance(spiders_list, list):
-        raise_shub_image_info_error('spiders section must be a list')
+    if not isinstance(project_type, string_types):
+        raise_shub_image_info_error('"project_type" key is required and must be a string')
 
-    project_type = metadata.get('project_type')
+    spiders_list = metadata.get('spiders')
+    if not isinstance(spiders_list, list):
+        raise_shub_image_info_error('"spiders" key is required and must be a list')
     spiders, scripts = [], []
     for name in spiders_list:
         if not (name and isinstance(name, string_types)):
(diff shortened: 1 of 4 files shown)


shub/image/list.py
```diff
@@ -81,7 +81,10 @@ def list_cmd(image_name, project, endpoint, apikey):
             click.echo(logs)
             raise ShubException('Container with list cmd exited with code %s' % exit_code)
         logs = logs.decode('utf-8') if isinstance(logs, binary_type) else logs
-        return {'spiders': utils.valid_spiders(logs.splitlines())}
+        return {
+            'project_type': 'scrapy',
```
vshlapakov: We haven't released non-scrapy deploys yet, so probably it's fine to assume `scrapy` project type if `shub-image-info` wasn't found.

shub/image/list.py
```diff
@@ -140,12 +143,17 @@ def raise_shub_image_info_error(error):
     if not isinstance(spiders_list, list):
         raise_shub_image_info_error('spiders section must be a list')
 
+    project_type = metadata.get('project_type')
     spiders, scripts = [], []
     for name in spiders_list:
         if not (name and isinstance(name, string_types)):
             raise_shub_image_info_error("spider name can't be empty or non-string")
-        scripts.append(name[3:]) if name.startswith('py:') else spiders.append(name)
+        if project_type == 'scrapy' and name.startswith('py:'):
```
vshlapakov: My guess is that scripts are something scrapy-specific, so it looks valid to me to add an additional condition here.

shub/image/list.py
```diff
@@ -140,12 +143,17 @@ def raise_shub_image_info_error(error):
     if not isinstance(spiders_list, list):
         raise_shub_image_info_error('spiders section must be a list')
 
+    project_type = metadata.get('project_type')
```
jdemaeyer: This allows breaking the contract by not supplying a project type. Will we allow `None` on the server side? Otherwise maybe we should default to `scrapy` or raise an error here (as we declared `project_type` to be required [in #280](https://github.com/scrapinghub/shub/pull/280/files#diff-eac378ae7fa584fd97b26cf7f2799382R36))

chekunkov: @jdemaeyer we added server-side validation that'll return an error if project_type is missing or invalid. is there any good reason to duplicate server-side validation in shub?

jdemaeyer: Validating the variable type is different from validating the variable value. IMO it is the client's responsibility to not make malformed requests (and the server's responsibility to reject correctly formed but disallowed values). Checking the type of `project_type` would be consistent with how we handle `spiders` and its entries (checking that the type is list or string, respectively).

More importantly, if we allow `None` here then `shub image test` will pass for an image that breaches the contract.

vshlapakov: @chekunkov @jdemaeyer Please take a qquick look, I've updated code and added some tests.

PR #290, opened 2017-07-07
shub-image-info output should be a single line

chekunkov:
We assume that when parse command output but docs suggest multiline output

### DIFF
--- a/docs/custom-images-contract.rst
+++ b/docs/custom-images-contract.rst
@@ -37,12 +37,8 @@ Contract statements
 
 .. code-block:: bash
 
-   docker run myscrapyimage shub-image-info
-
-    {
-        "project_type": "casperjs",
-        "spiders": ["spiderA", "spiderB"]
-    }
+    docker run myscrapyimage shub-image-info
+    {"project_type": "casperjs", "spiders": ["spiderA", "spiderB"]}
 
 .. note::
 


jdemaeyer:
Why (where) do we expect single-line output? The multi-line output as suggested by the docs is valid JSON and as such we should be able to parse it properly.

chekunkov:
Okay, PR description is badly worded. Here's the problem - we parse output from the stdout, which may also contain warnings or some info messages. If we assume that `shub-image-info` output is one line - we can iterate over lines of stdout and first successfully parsed json object is what we are looking for. If we allow multiline JSON object - parsing becomes harder and prone to errors - we need to determine beginning and end of the object in the output to parse it reliably - it may not always be first `{` and last `}`.

IMO that output is not intended to be human readable and if someone wants to read it he may use `jq` or similar tool to prettify output - so we can restrict output to be single line.

Any good reason to allow multiline output?

jdemaeyer:
> If we assume that shub-image-info output is one line - we can iterate over lines of stdout and first successfully parsed json object is what we are looking for.

Is that happening in the Kumo builder? [In shub](https://github.com/scrapinghub/shub/blob/master/shub/image/list.py#L139) it looks like we're always trying to parse the complete output.

> Any good reason to allow multiline output?

I was slightly afraid that the default JSON output libraries in some languages may print multi-line JSON, although I have now checked Ruby, PHP, Node.JS, and Python, and none of them do so, so maybe it's not an issue.

> IMO that output is not intended to be human readable and if someone wants to read it he may use jq or similar tool to prettify output - so we can restrict output to be single line.

I agree that human-readability is not a priority. It is the idea of rejecting perfectly valid JSON that I'm not really fond of. As a compromise, maybe we can try to parse the whole output first and only if that fails fall back to the "first valid line" approach?

In any case, I think we can merge this PR as single-line output is certainly a more realistic example.

chekunkov:
> As a compromise, maybe we can try to parse the whole output first and only if that fails fall back to the "first valid line" approach?

I'm okay with that

PR #292, opened 2017-07-21
Fix shub-image-info fallback

chekunkov:
docker.errors.NotFound is raised when Docker API returns 404, while in case of missing executable Docker API returns 500

### DIFF
--- a/shub/image/list.py
+++ b/shub/image/list.py
@@ -117,9 +117,15 @@ def _run_cmd_in_docker_container(image_name, command, environment):
         raise ShubException("Create container error:\n %s" % container)
     try:
         client.start(container)
-    except docker.errors.NotFound:
-        # return a proper exit code if executable is not found
-        return 127, None
+    except docker.errors.APIError as e:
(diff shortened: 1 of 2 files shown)


shub/image/list.py
```diff
@@ -117,9 +117,15 @@ def _run_cmd_in_docker_container(image_name, command, environment):
         raise ShubException("Create container error:\n %s" % container)
     try:
         client.start(container)
-    except docker.errors.NotFound:
-        # return a proper exit code if executable is not found
-        return 127, None
+    except docker.errors.APIError as e:
+        explanation = e.explanation or ''
+        if 'executable file not found' in explanation:
```
vshlapakov: Just to confirm, is there same message on `docker.errors.NotFound` cases?

chekunkov: @vshlapakov tbh I don't think 404 is ever raised in that case. I've seen only 500.

jdemaeyer:
Thanks Alex!

PR #295, opened 2017-07-31
Py3 compatibility fixes for image-list cmd

vshlapakov:
Closes #288, closes #296

PR #296, opened 2017-07-31
Handle Bytes input to JSON in Python 3

cathalgarvey:
I encountered a bug in Python 3.5.2 where some inputs to `json.loads` within `shub/image/list.py` were being passed as `bytes`, which lead to crashing. The fix is inelegant, but it passes tests on 2.7 and 3.5.

### DIFF
--- a/shub/image/list.py
+++ b/shub/image/list.py
@@ -3,6 +3,7 @@ import json
 import click
 import docker
 import requests
+import six
 from six import binary_type
 from six import string_types
 from six.moves.urllib.parse import urljoin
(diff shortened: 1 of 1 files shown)


cathalgarvey:
I call shenanigans on the failing test above: It's a travis test on Python 3.3 that errors out with: "ERROR: InterpreterNotFound: python3.3".

jdemaeyer:
Thanks alot Cathal, and sorry about the untimely reply. :/

@vshlapakov looks like this addresses the same issue as #295. Which do you prefer? I would rather go with #295 as it's closer to the root of the problem.

jdemaeyer:
> I call shenanigans on the failing test above: It's a travis test on Python 3.3 that errors out with: "ERROR: InterpreterNotFound: python3.3".

Indeed, if you rebase onto master it should go away.

vshlapakov:
@jdemaeyer +1, it addresses the same issue, I've just updated #295.

PR #301, opened 2017-10-25
Extract scripts if not found in shub-image-info

vshlapakov:
According to the last changes in [SH contract](https://shub.readthedocs.io/en/stable/custom-images-contract.html), there's `shub-image-info` command responsible for getting metadata and spiders list from a custom image. And we still support scripts in SH, so there's a need to get it in a similar way. However we can't implement getting scripts in default `shub-image-info` implementation in `scrapinghub-entrypoint-scrapy` (because it knows nothing about main project python egg and its internals), but we still should provide some workaround to make it work with custom images.

My proposal here is to always try to extract scripts if they're not found in `shub-image-info` output. Additional time is small, but this trick will help to solve the problem and bring scripts to custom images with `shub-image-info`.

Please, review.

### DIFF
--- a/shub/image/deploy.py
+++ b/shub/image/deploy.py
@@ -196,7 +196,7 @@ def _prepare_deploy_params(project, version, image_name, endpoint, apikey,
                            username, password, email):
     # Reusing shub.image.list logic to get spiders list
     metadata = list_mod.list_cmd(image_name, project, endpoint, apikey)
-    if 'scripts' not in metadata:
+    if not metadata.get('scripts'):
         metadata['scripts'] = _extract_scripts_from_project()
     params = {
         'project': project,


jdemaeyer:
Hey @vshlapakov, unfortunately my knowledge on custom images is still limited. The change looks like this to me:

* Before, we would check for project scripts if `shub-image-info` did not return any `scripts` field,
* Now, we _force_ getting the project scripts even if `shub-image-info` explicitly returned `scripts: []`.

Returning a `scripts` field is not part of the Custom Images contract. Why would `shub-image-info` return `scripts: []` when it actually hasn't checked for the scripts, instead of just leaving out this field? Is this how it's implemented in some base images we provide?

.travis.yml
```diff
@@ -3,13 +3,14 @@ python: 3.5
...
   - TOX_ENV=py27
-  - TOX_ENV=py33
   - TOX_ENV=py34
   - TOX_ENV=py35
   - TOX_ENV=freeze
 
 matrix:
   include:
+    - python: 3.3
+      env: TOX_ENV=py33
```
jdemaeyer: I shamelessly stole this for #303 ;)

vshlapakov:
@jdemaeyer These are all good questions, let me try to explain it and think on it together: 
- first of all, it's true that `scripts` is not a part of the contract
- therefore `shub-image-info` doesn't return this field at all (not before, not now, not in future)
- but we consider entries in its `spiders` list as scripts if it starts with `py:`
- so we split the list https://github.com/scrapinghub/shub/blob/master/shub/image/list.py#L155-L162
- and now it always returns `spiders` and `scripts` from the list command

I know, the decision about considering spiders starting from `py:` is questionable, but we had to do it for backward compatibility: to allow using scripts on platform by non-scrapy/non-python projects for some time until we support scripts (hoping it's not for a long time), that's there's no an explicit `scripts` field in `shub-image-info` contract.

So it becomes tricky due to different ways of deploy and compatibility issues, and deploying with `shub` is special because standard `shub-image-info` provided by `scrapinghub-entrypoint-scrapy` knows nothing about scripts and returns only spiders, and the only way to get scripts is to parse setup.py like it's done now. 

Let me know if there're still any uncertainties, I can also explain logic and differences behind deploys from Github if you're interested, probably it will help to structurize all the hacks and tricks and prepare some good internal doc.

jdemaeyer:
Hi @vshlapakov, thanks for the explanation.

I still struggle to follow, though. All that this PR does it to enforce script auto-detection even if `scripts: []` was returned by `shub-image-info`. (In contrast to the current implementation which auto-detects only if the `scripts` field is missing entirely.)

For any non-Scrapy custom image project, it seems to me that it would be easy to tweak the output of `shub-image-info` to return a `scripts` field, or to leave it out if you want script auto-detection. For any Scrapy project, or project based on our Scrapy image, we _know_ that there will be no `scripts` field, and that our current check (`if 'scripts' not in metadata`) will always trigger auto-detection.

Could you maybe explain a use case where `shub-image-info` would return `scripts: []` even though the user wants auto-detection?

vshlapakov:
@jdemaeyer First of all let's distinguish 2 different things:
- `shub-image-info` - a part of SH contract, its output contains only a strict fields set (so far only `project_type` and `spiders`)
- `shub.image.list_cmd` - a part of internal `shub` tool logic, its output can contain [anything](https://github.com/scrapinghub/shub/blob/master/shub/image/list.py#L155-L167) that could be useful for other commands, and in this case an additional `scripts` field is "calculated" from `spiders` field of `shub-image-info`.

> All that this PR does it to enforce script auto-detection even if `scripts: []` was returned by `shub-image-info`

`shub-image-info` never returned `scripts` field and it's not going to be changed in the future. Its output can contain only `spiders` field which can be interpreted as set of a spiders and maybe scripts. So, fixing your statement: ```all that this PR does it to enforce script auto-detection even if `scripts: []` was returned by `shub.image.list_cmd```

> it would be easy to tweak the output of `shub-image-info` to return a `scripts` field

You're right, but as far as I remember our discussions with @chekunkov and @dangra, we don't want to add the field to the contract: we're going to get rid of the whole scripts concept in the future and adding it to the contract now will only confuse people later. For me all the current logic related with scripts is directed only toward covering backward-compatibility issues. 
 
Anyway I have to admit that your version with auto-detection sounds simple and solid. Requesting for @dangra thoughts on that: I wonder what are our current plans on dropping scripts and if there're any estimates. If it's postponed for undefined time, maybe it's simpler to extend current contract, include optional `scripts` field to support current platform needs and deprecate it later when we're ready. Because as of now I see too many tricks and work-arounds to make it work :thinking:

vshlapakov:
@jdemaeyer Hey, any estimates it could be merged?

vshlapakov:
Alright, merging this as-is for now, let's get back to this discussion later when we have more resources.

PR #302, opened 2017-11-08
Allow custom Dockerfile names

jesuslosada:
This modification allows to pass a custom Dockerfile name instead of assuming the default value ('Dockerfile') when building images. This is useful when you want to manage different Dockerfiles in the same repo.

### DIFF
--- a/shub/image/build.py
+++ b/shub/image/build.py
@@ -39,18 +39,20 @@ BUILD_SUCCESS_REGEX = re.compile(r'Successfully built ([0-9a-f]+)')
               help="stream build logs to console")
 @click.option("-V", "--version", help="release version")
 @click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
-def cli(target, debug, verbose, version, skip_tests):
-    build_cmd(target, version, skip_tests)
+@click.option("-f", "--file", "filename", default='Dockerfile',
+              help="Name of the Dockerfile (Default is 'PATH/Dockerfile')")
+def cli(target, debug, verbose, version, skip_tests, filename):
+    build_cmd(target, version, skip_tests, filename=filename)
 
 
-def build_cmd(target, version, skip_tests):
+def build_cmd(target, version, skip_tests, filename='Dockerfile'):
     config = load_shub_config()
     create_scrapinghub_yml_wizard(config, target=target, image=True)
     client = utils.get_docker_client()
     project_dir = utils.get_project_dir()
     image = config.get_image(target)
     image_name = utils.format_image_name(image, version)
-    if not os.path.exists(os.path.join(project_dir, 'Dockerfile')):
+    if not os.path.exists(os.path.join(project_dir, filename)):
         raise shub_exceptions.NotFoundException(
             "Dockerfile is not found and it is required because project '{}' is configured "
             "to deploy Docker images. Please add a Dockerfile that will be used to build "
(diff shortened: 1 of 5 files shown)


shub/image/build.py
```diff
@@ -39,18 +39,21 @@
...
-def build_cmd(target, version, skip_tests):
+def build_cmd(target, version, skip_tests, filename=None):
     config = load_shub_config()
     create_scrapinghub_yml_wizard(config, target=target, image=True)
     client = utils.get_docker_client()
     project_dir = utils.get_project_dir()
     image = config.get_image(target)
     image_name = utils.format_image_name(image, version)
-    if not os.path.exists(os.path.join(project_dir, 'Dockerfile')):
+    dockerfile = filename or 'Dockerfile'
```
jdemaeyer: (Together with the default value for the click option we could set `filename='Dockerfile'` in the parameter list above and avoid this line.)

shub/image/build.py
```diff
@@ -39,18 +39,21 @@
               help="stream build logs to console")
 @click.option("-V", "--version", help="release version")
 @click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
-def cli(target, debug, verbose, version, skip_tests):
-    build_cmd(target, version, skip_tests)
+@click.option("-f", "--file", "filename",
+              help="Name of the Dockerfile (Default is 'PATH/Dockerfile')")
```
jdemaeyer: We can use `default='Dockerfile'` in this option definition (it'll also show up nicely in the help string then).

shub/image/build.py
```diff
@@ -61,7 +64,12 @@ def build_cmd(target, version, skip_tests):
     else:
         build_progress_cls = _BuildProgress
     click.echo("Building {}.".format(image_name))
-    events = client.build(path=project_dir, tag=image_name, decode=True)
+    events = client.build(
+        path=project_dir,
+        tag=image_name,
+        decode=True,
+        dockerfile=filename
```
jdemaeyer: Should this be `dockerfile=dockerfile`? I guess `client.build` will default to `Dockerfile` anyways, but still, I would rather not use two different variables for the same thing.

jdemaeyer: Or is there some "smart Dockerfile finding" that we would miss out on when supplying `'Dockerfile'` instead of `None`?

jesuslosada: I'm not 100% sure about the logic applied by Docker, but I was using None instead of 'Dockerfile' because None is the default value in docker-py's build method: https://github.com/docker/docker-py/blob/master/docker/api/build.py#L16

Let me know if you prefer to change that.

jdemaeyer: Hm, that line is over two years old (a very long time before we even introduced custom images for Scrapy Cloud), and the error message would be quite cryptic to `shub` users anyways. I'd prefer to enforce `docker-py>=1.17` in our `setup.py` and `requirements.in` instead.

I realized we're not using any smart Dockerfile-finding by `client.build()` anyways because we explicitly enforce the existence of "project_dir/Dockerfile" a few lines above. Given all this I'd prefer to set defaults in the click option and the function parameter list, and avoid the dual variables.

vshlapakov:
@jdemaeyer @jesuslosada No objections, that will be a nice addition to the command :+1: +1 to enforce `docker-py>=1.17`, rely on default value provided by click and use a single variable for Dockerfile path as we anyway check for its existence before passing it further.

vshlapakov:
@jesuslosada Btw to fix the Travis build errors please rebase the PR, the fix is already in master branch.

vshlapakov:
Hey @jesuslosada I'm taking ownership of the project, could you please rebase the PR?

jesuslosada:
Hi @vshlapakov . Sorry for the delay. I've updated the PR.

vshlapakov:
@jesuslosada No worries, thanks a lot for your help :+1:

PR #304, opened 2017-11-28
Make `get_docker_client` compatible with old and new versions of docker-py

arruda:
The docker Client class is now called DockerClient in newer versions of the docker python
package.

There's also the problem that the `docker-py` no longer is called docker-py, only `docker`.
So now if you install have the docker-py installed, it will obfuscurate the new version of the package,
and you are advised to remove it and make `pip install docker` again.

That's why I also changed the package name in the requirements

About the tests, I would require some help in this, since I'm not very good with the Tox envs.
But I know that a solution for this tests could be done by adding a env for two versions of the docker-py package (the old, named `docker-py` and the new that's known as `docker`).

This way the tests for each python version would also be tested in combination with the two possible docker-py versions. Ex: `py27-dockerpyold`; `py27-dockerpynew`, `py35-dockerpyold`; `py35-dockerpynew`, etc...

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-backports.ssl-match-hostname==3.5.0.1  # via docker-py, websocket-client
+backports.ssl-match-hostname==3.5.0.1  # via docker-py
 click==6.6
 docker-py==1.10.6
 docker-pycreds==0.2.1     # via docker-py
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,15 +4,15 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-backports.ssl-match-hostname==3.5.0.1  # via docker, websocket-client
+backports.ssl-match-hostname==3.5.0.1  # via docker-py, websocket-client
 click==6.6
-docker-pycreds==0.2.1     # via docker
-docker==2.6.1
-ipaddress==1.0.17         # via docker
+docker-py==1.10.6
+docker-pycreds==0.2.1     # via docker-py
+ipaddress==1.0.17         # via docker-py
 pyyaml==3.12
 requests==2.10.0
 retrying==1.3.3
 scrapinghub==1.9.0
 six==1.10.0
 tqdm==4.11.2
-websocket-client==0.37.0  # via docker
+websocket-client==0.37.0  # via docker-py
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,15 +4,15 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-backports.ssl-match-hostname==3.5.0.1  # via docker-py
+backports.ssl-match-hostname==3.5.0.1  # via docker, websocket-client
 click==6.6
-docker-py==1.10.6
-docker-pycreds==0.2.1     # via docker-py
-ipaddress==1.0.17         # via docker-py
+docker-pycreds==0.2.1     # via docker
+docker==2.6.1
+ipaddress==1.0.17         # via docker
 pyyaml==3.12
 requests==2.10.0
 retrying==1.3.3
 scrapinghub==1.9.0
 six==1.10.0
 tqdm==4.11.2
-websocket-client==0.37.0  # via docker-py
+websocket-client==0.37.0  # via docker
--- a/shub/image/utils.py
+++ b/shub/image/utils.py
@@ -75,9 +75,10 @@ def get_docker_client(validate=True):
     except ImportError:
         raise ImportError('You need docker python package installed for the cmd')
 
-    # added for compatiblity with old docker-py
+    # docker-py (legacy)
     if hasattr(docker, 'Client'):
         docker_client_cls = docker.Client
+    # docker >= 2.0
     else:
         docker_client_cls = docker.APIClient
 
@@ -95,7 +96,6 @@ def get_docker_client(validate=True):
             assert_hostname=False)
         docker_host = docker_host.replace('tcp://', 'https://')
     version = os.environ.get('DOCKER_VERSION', DEFAULT_DOCKER_VERSION)
-
     client = docker_client_cls(base_url=docker_host,
                                version=version,
                                tls=tls_config)
--- a/shub/image/utils.py
+++ b/shub/image/utils.py
@@ -79,7 +79,7 @@ def get_docker_client(validate=True):
     if hasattr(docker, 'Client'):
         docker_client_cls = docker.Client
     else:
-        docker_client_cls = docker.DockerClient
+        docker_client_cls = docker.APIClient
 
     docker_host = os.environ.get('DOCKER_HOST')
     tls_config = None
--- a/shub/image/utils.py
+++ b/shub/image/utils.py
@@ -73,7 +73,13 @@ def get_docker_client(validate=True):
     try:
         import docker
     except ImportError:
-        raise ImportError('You need docker-py installed for the cmd')
+        raise ImportError('You need docker python package installed for the cmd')
+
+    # added for compatiblity with old docker-py
+    if hasattr(docker, 'Client'):
+        docker_client_cls = docker.Client
+    else:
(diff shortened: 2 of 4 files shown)


pablohoffman:
always nice to see `shub` contributions from hubbers!

arruda:
:v:

jdemaeyer:
Hi Felipe, thanks for the PR!

`docker.Client` has actually been renamed to `docker.APIClient`. `DockerClient` is a new, more user-friendly (but similar) API that has been introduced in [docker 2.0](https://github.com/docker/docker-py/releases/tag/2.0.0). To avoid any unexpected pitfalls we should probably use `APIClient` (and do a refactor with `DockerClient` later).

The renaming of the docker package is quite tricky for us:

First, the `requirements.in/txt` files are only used for tests and building binaries, the actual packages that will be pulled on `pip install shub` are [listed in setup.py](https://github.com/scrapinghub/shub/blob/master/setup.py#L22). So we'd still be installing `docker-py` on `pip install shub`.

Now, we can easily fix this by exchanging `docker-py` for `docker` in setup.py as well, but there's a big problem with that: it may break user environments. Assume they have `docker-py` installed (for example after installing our current shub release) and have a script that uses `from docker import Client`. Now we update setup.py and they upgrade shub, which pulls `docker`. Suddenly, their script stops working:

```
$ python -m virtualenv .venv
Using base prefix '/usr'
New python executable in /home/jakob/playground/dockerblah/.venv/bin/python
Installing setuptools, pip, wheel...done.
$ source .venv/bin/activate
$ pip install docker-py
[...]
Successfully installed certifi-2017.11.5 chardet-3.0.4 docker-py-1.10.6 docker-pycreds-0.2.1 idna-2.6 requests-2.18.4 six-1.11.0 urllib3-1.22 websocket-client-0.44.0

$ python -c "from docker import Client"

$ pip install docker
[...]
Successfully installed docker-2.6.1

$ python -c "from docker import Client"
  File "<string>", line 1, in <module>
ImportError: cannot import name 'Client'
```

(The same is true with our current setup.py if the user had a script importing `docker.Dockerclient` before installing shub, but I guess it's safe to assume that all current production setups use `docker-py` or else they wouldn't be able to use shub.)

I'm not sure how to solve this. We cannot dynamically generate `install_requires` (based on `docker.__version__` if available) as it is evaluated at wheel creation time and not at install time, and there is no [environment marker](https://www.python.org/dev/peps/pep-0508/#environment-markers) that would allow us to check for an installed docker package. Maybe we can support the newer docker package via `extras_require={'docker': ['docker']}`? But then we might be pulling _both_ `docker` and `docker-py` and pip's implementation would define which shadows which (I assume).

I doubt we're the only ones with this problem but couldn't find any solutions from other projects so far. What do you think @arruda @vshlapakov?

arruda:
About the docker-py and docker namespace issue, what I've been thinking is:
We cannot remove the old docker-py even if we knew it was there (I believe it's possible to know that it's there). 
But maybe we should follow the lead on the docker team, with the docker-compose tool:
if you have the docker-py, and then you install docker-compose, then it will install docker-compose with the docker as  a dependency.
And when you try to run something it will give you a warning about the this:
```
$ docker-compose ps
WARNING: Dependency conflict: an older version of the 'docker-py' package may be polluting the namespace. If you're experiencing crashes, run the following command to remedy the issue:
pip uninstall docker-py; pip uninstall docker; pip install docker
```

I'm not even sure if we should remove docker-py even if we where able to :thinking:
Another idea, but I'm not sure it's 100% safe, is to have this warning/exception in the installation (setup.py).
So it wouldn't install anything while you have this messy dependency (docker-py) installed. The problem is that you might have `shub` and `docker-py` in the same requirements, been installed together, then this wouldn't get it... that's why I think the warning solution inside the shub run, is actually better.

arruda:
About the APIClient, I'll change that :+1:

arruda:
I changed the APIClient and also the setup.py for now.

jdemaeyer:
> I'm not even sure if we should remove docker-py even if we where able to 🤔

For sure not. As I outlined above, removing `docker-py` can break user environments/scripts. I think the only way we can move to `docker` as our main dependency is in a backwards-compatibility-breaking major release. (Which I would actually like to do relatively soon, so that'd be our chance.)

> Another idea, but I'm not sure it's 100% safe, is to have this warning/exception in the installation (setup.py).
> So it wouldn't install anything while you have this messy dependency (docker-py) installed.

During a wheel installation, setup.py is never executed (in fact it is not even part of the wheel), so nothing we can do in there :/

The more I think about this the more I think the only thing we can do in the 2.x branch is allow users to manually upgrade to `docker` without breaking shub.

arruda:
:thinking: @jdemaeyer, I think you're right. I mean, this package problem with docker-py/docker is a real mess.
If we keep the docker-py as the dependency, and keep the code compatible with the new one, shouldn't create any new problem. If the user manually update the docker it won't be a problem. The only problem that I see in this, is that if the user run a `pip install -U shub` at some point afterwards, then it will again install the docker-py lib, and will mess with the user env again =/

But the thing is, I don't think there's a good way around this, and by doing keeping the old version and adding compatibility to the new, we can, somehow, postpone the dependency problem to another time, maybe until we figure out a better solution.

Meanwhile, maybe, the docker team might do something about this name conflict.

arruda:
Hummm, there was a bizarre CI error when fetching something related to homebrew in the OSX freeze job:https://travis-ci.org/scrapinghub/shub/jobs/308491248
I can't re-run that job, maybe someone else have access to do that. @jdemaeyer ?

shub/image/utils.py
```diff
@@ -73,7 +73,13 @@ def get_docker_client(validate=True):
...
         import docker
     except ImportError:
-        raise ImportError('You need docker-py installed for the cmd')
+        raise ImportError('You need docker python package installed for the cmd')
+
+    # added for compatiblity with old docker-py
+    if hasattr(docker, 'Client'):
+        docker_client_cls = docker.Client
+    else:
+        docker_client_cls = docker.APIClient
```
jdemaeyer: Can you make the comment a little more detailed? E.g. remove the comment at the top and add `# docker-py (legacy)` and `# docker >= 2.0` into the two branches.

shub/image/utils.py
```diff
@@ -89,9 +95,10 @@ def get_docker_client(validate=True):
             assert_hostname=False)
         docker_host = docker_host.replace('tcp://', 'https://')
     version = os.environ.get('DOCKER_VERSION', DEFAULT_DOCKER_VERSION)
-    client = docker.Client(base_url=docker_host,
-                           version=version,
-                           tls=tls_config)
+
```
jdemaeyer: (nitpick) no extra blank lines please :)

vshlapakov:
Sorry gentlemen, I'm a bit late to the party, but generally speaking I have nothing to add to what was already said: let's keep oldy `docker-py` in requirements and add support for newer `docker` client if it's installed - that's the best we can do right now, and schedule `docker-py`->`docker` migration for next major `shub` release.

requirements.txt
```diff
@@ -4,7 +4,7 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-backports.ssl-match-hostname==3.5.0.1  # via docker-py
+backports.ssl-match-hostname==3.5.0.1  # via docker-py, websocket-client
```
vshlapakov: Very minor thing, but don't think it's related with other changes in the PR:
it only means that [your python version is older than `2.7.10`](https://github.com/websocket-client/websocket-client/blob/master/setup.py#L15-L21).

arruda: You're right, I'll fix this.

shub/image/utils.py
```diff
@@ -73,7 +73,14 @@ def get_docker_client(validate=True):
     try:
         import docker
     except ImportError:
-        raise ImportError('You need docker-py installed for the cmd')
+        raise ImportError('You need docker python package installed for the cmd')
```
vshlapakov: Rollback the comment to docker-py?

jdemaeyer:
Thanks Felipe!

PR #309, opened 2018-04-20
Replace doc.sh.com links with actual ones

vshlapakov:
Some docs pointing to `docs.sh.com` are outdated, let's update it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,21 @@
 from __future__ import absolute_import
+import io
+import os
 from setuptools import setup, find_packages
+
+
+about = {}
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -6,7 +6,7 @@
     name='shub',
     version='2.7.0',
     packages=find_packages(exclude=('tests', 'tests.*')),
-    url='http://doc.scrapinghub.com/shub.html',
+    url='https://shub.readthedocs.io',
```
jdemaeyer: Maybe we can have one central place where we define all the docs links? E.g. in `shub/__init__.py`, and then always import from there?

PR #310, opened 2018-04-20
Schedule: add support for priority, units, tags

### LINKED ISSUE #308: Complete schedule API
Scrapy Cloud's `run.json` API endpoint supports specifying the number of SC units to use, and adding tags to the jobs. These options should also be available in `shub schedule`.

elacuesta:
My two cents to fix #308

### DIFF
--- a/shub/schedule.py
+++ b/shub/schedule.py
@@ -38,22 +38,31 @@ Similarly, job-specific settings can be supplied through the -s option:
 """
 
 SHORT_HELP = "Schedule a spider to run on Scrapy Cloud"
+DEFAULT_PRIORITY = 2
+DEFAULT_UNITS = 1
 
 
 @click.command(help=HELP, short_help=SHORT_HELP)
 @click.argument('spider', type=click.STRING)
 @click.option('-a', '--argument',
-              help='spider argument (-a name=value)', multiple=True)
+              help='Spider argument (-a name=value)', multiple=True)
 @click.option('-s', '--set',
-              help='job-specific setting (-s name=value)', multiple=True)
-def cli(spider, argument, set):
+              help='Job-specific setting (-s name=value)', multiple=True)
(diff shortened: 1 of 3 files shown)


shub/schedule.py
```diff
@@ -43,17 +43,24 @@
 @click.command(help=HELP, short_help=SHORT_HELP)
 @click.argument('spider', type=click.STRING)
 @click.option('-a', '--argument',
-              help='spider argument (-a name=value)', multiple=True)
+              help='Spider argument (-a name=value)', multiple=True)
 @click.option('-s', '--set',
-              help='job-specific setting (-s name=value)', multiple=True)
-def cli(spider, argument, set):
+              help='Job-specific setting (-s name=value)', multiple=True)
+@click.option('-p', '--priority', default=2,
```
vshlapakov: Let's add int type check for `priority`/`units` options?

shub/schedule.py
```diff
@@ -69,12 +76,15 @@ def cli(spider, argument, set):
 
 
 def schedule_spider(project, endpoint, apikey, spider, arguments=(),
-                    settings=()):
+                    settings=(), priority=2, units=1, tag=()):
```
vshlapakov: I'd put the default values into constants, like `DEFAULT_PRIORITY` etc, to define it only once.

shub/schedule.py
elacuesta: `int` is assumed because of the default value, but I agree that _explicit is better than implicit_

elacuesta:
Hey Viktor, thanks for the review. I fixed a few small things but I still need to update the tests, I'll do it shortly.
Thanks again for your time!

elacuesta:
Squashed, many thanks @vshlapakov 🙌

vshlapakov:
@elacuesta :+1: It's you who did all the work :slightly_smiling_face:

Issue #317, opened 2018-06-04
Cannot deploy from Pipfile with VCS package


hcoura:
@vshlapakov You said you were releasing a new version this week, when I was using master at a project of mine I ran into this problem.

I have already found the reason and I am working on a PR to be submitted soon. I just opened this issue to you know about it and may decide to hold on releasing a new version.

PR #318, opened 2018-06-04
Ignore hashes on Pipfile.lock to play nice with vcs packages

hcoura:
Fixes https://github.com/scrapinghub/shub/issues/317

VCS packages don't generate hashes, and `convert_deps_to_pip` generates requirements with hashes.

When we run `pip install -r requirements.txt` with a `requirements.txt` like:
```
package==1.0 --hash=hash
git+https://github... # note that it doesn't have the --hash
```
It fails since the vcs package can't be hashed

Thus we must delete these hashes from the object before giving it to `convert_deps_to_pip`

Btw, they do something similar in their own code: https://github.com/pypa/pipenv/blob/master/pipenv/utils.py#L793

### DIFF
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -174,6 +174,13 @@ def _get_pipfile_requirements():
             deps = json.load(f)['default']
     except IOError:
         raise ShubException('Please lock your Pipfile before deploying')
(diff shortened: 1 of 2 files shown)


vshlapakov:
Amazing, thanks @hcoura :+1:

Issue #322, opened 2018-07-10
shub image push: wrong registry detection for Azure


vshlapakov:
When using a custom registry, image format is the following: `some-registry.domain/user/repo`, but it's different for Azure container registry where it's `some-azure-registry.io/repo`. As a result, login with Azure registry fails because it [detects](https://github.com/scrapinghub/shub/blob/85b112daf03cbdeabada4b25fb28136ee2d9c93a/shub/image/push.py#L75-L76) a registry incorrectly.
```
In [15]: image = 'some-azure-registry.io/repository:tag'
In [16]: len(image.split('/'))
Out[16]: 2
```

vshlapakov:
Fixed with https://github.com/scrapinghub/shub/pull/324, closing.

PR #324, opened 2018-07-11
Improve image registry parsing logic

vshlapakov:
Follow-up to #322.


vshlapakov:
Only tests for Python 3.3 fails because of https://github.com/pypa/setuptools/pull/1342, I think it should be fine to drop support for Python 3.3 in the future versions, will address it in a separate PR.

PR #327, opened 2018-07-16
Remove use of Py3.7 reserved word

### LINKED ISSUE #325: Python3.7 compatibility
With the new release of Python 3.7. `async` is a keyword, so the use of `async` as a variable results in `SyntaxError: invalid syntax`. A quick search in the code reveals `async` was used in `image/deploy.py` and `image/upload.py`.

rennerocha:
`async` is a reserved word in Python 3.7 (https://docs.python.org/3.7/reference/lexical_analysis.html#keywords) and `shub` uses it as a variable.

This PR closes #325 and remove the use of this reserved word, avoiding SyntaxError when running in Python 3.7.

### DIFF
--- a/shub/image/deploy.py
+++ b/shub/image/deploy.py
@@ -55,16 +55,16 @@ Does a simple POST request to Dash API with given parameters
 @click.option("--email", help="docker registry email")
 @click.option("--apikey", help="SH apikey to use built-in registry")
 @click.option("--insecure", is_flag=True, help="use insecure registry")
-@click.option("--async", is_flag=True, help="[DEPRECATED] enable asynchronous mode",
+@click.option("--async", "_async", is_flag=True, help="[DEPRECATED] enable asynchronous mode",
               callback=utils.deprecate_async_parameter)
 def cli(target, debug, verbose, version, username, password, email,
-        apikey, insecure, async):
+        apikey, insecure, _async):
     deploy_cmd(target, version, username, password, email,
-               apikey, insecure, async)
+               apikey, insecure, _async)
 
 
 def deploy_cmd(target, version, username, password, email,
-               apikey, insecure, async):
+               apikey, insecure, _async):
     config = load_shub_config()
     target_conf = config.get_target_conf(target)
     endpoint, target_apikey = target_conf.endpoint, target_conf.apikey
@@ -99,7 +99,7 @@ def deploy_cmd(target, version, username, password, email,
     click.echo(
         "You can check deploy results later with "
         "'shub image check --id {}'.".format(status_id))
-    if async:
+    if _async:
         return
     if utils.is_verbose():
         deploy_progress_cls = _LoggedDeployProgress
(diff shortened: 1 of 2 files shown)


lucywang000:
The tests failed on py3.3, because tox has dropped support for python 3.3 (which is obsolete now). Maybe we can remove the support for py3.3 too in scrapy? Or we need to pin tox to to 2.9.x.

```
$ tox -e $TOX_ENV
  File "/home/travis/virtualenv/python3.3.6/bin/tox", line 7, in <module>
    from tox import cmdline
  File "/home/travis/virtualenv/python3.3.6/lib/python3.3/site-packages/tox/__init__.py", line 10, in <module>
    from pkg_resources import DistributionNotFound, get_distribution
  File "/home/travis/virtualenv/python3.3.6/lib/python3.3/site-packages/pkg_resources/__init__.py", line 85, in <module>
    raise RuntimeError("Python 3.4 or later is required")
RuntimeError: Python 3.4 or later is required
```

From tox changelog:
```
drop Python 2.6 and 3.3 support: setuptools dropped supporting these, and as we depend 
on it we’ll follow up with doing the same (use tox <= 2.9.1 if you still need this support)
- by @gaborbernat (#679)
```

shub/image/deploy.py
```diff
@@ -55,16 +55,16 @@
 @click.option("--email", help="docker registry email")
 @click.option("--apikey", help="SH apikey to use built-in registry")
 @click.option("--insecure", is_flag=True, help="use insecure registry")
-@click.option("--async", is_flag=True, help="[DEPRECATED] enable asynchronous mode",
+@click.option("--async", "_async", is_flag=True, help="[DEPRECATED] enable asynchronous mode",
```
dangra: I think the norm for using reserved keywords as variables names that are not private is to suffix (rather than prefix) with underscore.

rennerocha: I don't like to include underscore as well. I prefixed with underscore to follow the example in Click
http://click.pocoo.org/dev/options/#basic-value-options
```
# How to use a Python reserved word such as `from` as a parameter
@click.command()
@click.option('--from', '-f', '_from')
@click.option('--to', '-t')
def reserved_param_name(_from, to):
    click.echo('from %s to %s' % (_from, to))
```

dangra: that is an interesting conflict of standards! [PEP8 mentions](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) this about the topic:

> the following special forms using leading or trailing underscores are recognized (these can generally be combined with any case convention):
>
>   *  _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore.
>
 >  * single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.
>
>        Tkinter.Toplevel(master, class_='ClassName')

```

rennerocha:
Looking better at click's documentation and code, I found that this could be improved in their documentation:
https://github.com/pallets/click/pull/1073

Update this PR to follow PEP8 convention.

vshlapakov:
Thanks a lot @rennerocha! :tada:

jonaskkristensen:
I still get this error, running Python 3.7

PR #329, opened 2018-08-03
[MRG] do not set units=1 by default

lucywang000:
Dash supports per-spider units settings (e.g. spider foo.com have default units 2 for all its jobs, while bar.com have default 3 units for all its jobs), and in our projects (BV) we make extensive use of it.

After upgrading to shub 2.8.0, the jobs I scheduled always used 1 units and got OOM now and then due to that now and then.

I think the correct behavior is not to pass units=1 to the schedule call when the user doesn't specify a `--units` value.

### DIFF
--- a/shub/schedule.py
+++ b/shub/schedule.py
@@ -39,7 +39,6 @@ Similarly, job-specific settings can be supplied through the -s option:
 
 SHORT_HELP = "Schedule a spider to run on Scrapy Cloud"
 DEFAULT_PRIORITY = 2
-DEFAULT_UNITS = 1
 
 
 @click.command(help=HELP, short_help=SHORT_HELP)
@@ -50,7 +49,7 @@ DEFAULT_UNITS = 1
               help='Job-specific setting (-s name=value)', multiple=True)
 @click.option('-p', '--priority', type=int, default=DEFAULT_PRIORITY,
               help='Job priority (-p number). From 0 (lowest) to 4 (highest)')
-@click.option('-u', '--units', type=int, default=DEFAULT_UNITS,
+@click.option('-u', '--units', type=int, default=None,
               help='Amount of Scrapy Cloud units (-u number)')
 @click.option('-t', '--tag',
               help='Job tags (-t tag)', multiple=True)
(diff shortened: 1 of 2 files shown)


cathalgarvey:
@elacuesta - The blame suggests that you wrote the Default handling line prior to this change. The change looks clean at a code-level to me, but I'm worried we'll break someone else's workflow by removing the default. Can you offer advice on why the default behaviour was added, and if it will break things to remove it?

lucywang000:
> I'm worried we'll break someone else's workflow by removing the default

That makes sense. But still if someone relies on per-spider default units settings decides to upgrade to 2.7.0 to 2.8.0 then his/her workflow would be broken too.

lucywang000:
The tests failed at somewhere not related to this change:

```
_ TestDeployReqs.test_can_decompress_downloaded_packages_and_call_deploy_reqs __
self = <tests.test_deploy_reqs.TestDeployReqs testMethod=test_can_decompress_downloaded_packages_and_call_deploy_reqs>
    def test_can_decompress_downloaded_packages_and_call_deploy_reqs(self):
        requirements_file = self._write_tmp_requirements_file()
        with mock.patch('shub.utils.build_and_deploy_egg') as m:
            self.runner.invoke(
                deploy_reqs.cli,
                ('-r', requirements_file),
            )
>           self.assertEqual(m.call_count, 2)
E           AssertionError: 1 != 2
```

I'll take a look next Monday.

elacuesta:
I didn't take the per-project-units setting into account, sorry! I think this change makes a lot of sense.
There was no strong reason for adding the `DEFAULT_UNITS` value, it was just to provide a default value without hardcoding `1` everywhere in the file. I tested agains a project with no default unit number, that's why I didn't realize about this.

shub/schedule.py
```diff
@@ -50,7 +49,7 @@
               help='Job-specific setting (-s name=value)', multiple=True)
 @click.option('-p', '--priority', type=int, default=DEFAULT_PRIORITY,
               help='Job priority (-p number). From 0 (lowest) to 4 (highest)')
-@click.option('-u', '--units', type=int, default=DEFAULT_UNITS,
+@click.option('-u', '--units', type=int, default=None,
```
vshlapakov: Let's drop the `default` arg, an option takes `None` by default anyway

shub/schedule.py
```diff
@@ -78,16 +77,18 @@ def cli(spider, argument, set, priority, units, tag):
 
 
 def schedule_spider(project, endpoint, apikey, spider, arguments=(), settings=(),
-                    priority=DEFAULT_PRIORITY, units=DEFAULT_UNITS, tag=()):
+                    priority=DEFAULT_PRIORITY, units=None, tag=()):
     conn = Connection(apikey, url=endpoint)
     try:
+        kw = dict(x.split('=', 1) for x in arguments)
+        kw['add_tag'] = tag
```
vshlapakov: Let's pass it as a keyword arg as before, or there's a reason to change it?

vshlapakov:
> The tests failed at somewhere not related to this change

True, I've restarted tests for master, and it's [broken](https://travis-ci.org/scrapinghub/shub/jobs/405488804), I believe it shouldn't stop us from merging the PR. Unfortunately I don't have time to check & fix it during next days, but if you have some time to take a look - I appreciate it, thanks.

lucywang000:
@vshlapakov I addressed your comments, and also disabled the flaky test.

PR #336, opened 2018-10-30
Enclose example in code block

ttilberg:
Previously it was acting as a block quote.


vshlapakov:
Thanks @ttilberg! :raised_hands:

PR #337, opened 2018-10-30
Give information on using environment var: SHUB_APIKEY

ttilberg:
Fixes #335

Issue #335, opened 2018-10-30
Document SHUB_APIKEY


andrewbaxter:
A customer contacted us after spending some unspecified amount of time trying to get noninteractive deploy working for a CI workflow, and I ended up searching through other projects and then confirming in the shub sourcecode that `SHUB_APIKEY` exists and works.

I think a CI deploy is/should be a best practice since it provides more consistency in the deployed images and can enforce various deploy checks - to that end `SHUB_APIKEY` doesn't seem to be documented anywhere.

vshlapakov:
Addressed by https://github.com/scrapinghub/shub/pull/338, thanks guys @ttilberg @andrewbaxter!

PR #338, opened 2018-10-31
Describe using environment variables

vshlapakov:
Addresses #335 (extending #337).

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -11,7 +11,8 @@ shub is configured via two YAML files:
 
 * ``~/.scrapinghub.yml`` -- this file contains global configuration like
   your API key. It is automatically created in your home directory when you run
-  ``shub login``.
+  ``shub login``. You can also change the default location with an environment
+  variable, check an appropriate section below.
 * ``scrapinghub.yml`` -- this file contains local configuration like the
   project ID or the location of your requirements file. It is automatically
   created in your project directory when you run ``shub deploy`` for the first
@@ -120,6 +121,9 @@ On Linux-based systems::
 On Windows::
 
     SET SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
+
+You can also parametrize global ``scrapinghub.yml`` file location with
+``SHUB_GLOBAL_CONFIG`` environment variable (default ``~/.scrapinghub.yml``).
 
 When working with custom Docker images, please be aware that the tool relies
 on a set of standard ``DOCKER_`` prefixed environment variables:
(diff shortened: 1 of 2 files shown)


ttilberg:
Can you check the RST preview? Specifically in the quickstart, I don't think that's coming across as you expect. 

Additionally, I think that it is inconvenient to link out, isntead of just giving the environment var key `SHUB_APIKEY` as information up front, being that it's such little text as it is.

I agree that adding the information about the overall environment config is helpful.

vshlapakov:
@ttilberg Thanks a lot for checking!

> Can you check the RST preview?

The docs are not intended for reading on Github, it's compiled by Sphinx after merging and published on https://shub.readthedocs.io. You can test it locally by
1) `pip install Sphinx sphinx_rtd_theme`
2) `cd docs && make html`
2) start with `_build/html/index.html`

And as you can see, everything is fine there:
image
> instead of just giving the environment var key 

I think you are partly right here, but there are a few reasons why I insist to keep it in this way:

1) I believe the quickstart section must contain only minimal steps to start working with the tool. And despite using an environment variable could be helpful for automatization, it's not so convenient for people as using `shub login` because you have to set the env variable for each session while `shub login` sets it once and for all. 
2) A significant amount of `shub` users are using Windows and syntax is a bit different there, so it's not about a single tiny paragraph anymore.

Not saying that some non-technical people might be not well-familiar with environment variables and the paragraph might raise more questions than benefits. Summarizing, I'm convinced that it's an advanced topic and the best we could do here is to provide a reference to the appropriate section in the configuration document.

ttilberg:
> it's compiled by Sphinx after merging and published on https://shub.readthedocs.io.
Okay thanks, right. I'm completely unfamiliar with .rst, and was leaning on the preview to help me out. Looks good.

Regarding the Quickstart: I agree with the sentiment, it should be straightforward and not confusing. I'm also familiar with using env vars in both Windows and Linux and understand the syntax is different. So as an alternative, instead of giving the example of how to apply the env, I think it's worth at least noting what the env var is, along the lines of:


```
Alternatively, you can set your Scrapinghub API key as an environment variable (`SHUB_APIKEY`)...
```

Basically, we're already presenting information that such a thing exists, I think it's really important to at least make the information directly useful, particularly being in the quickstart, and particularly since it is also quite straight-forward.

I'm just looking at this from the perspective of it being the docs I wish I had read and printed out earlier this week. I would have found this immensely useful, and having the link to the previously undocumented information is immensely useful, but simply stating the env var we are teasing up front is even more so.

vshlapakov:
> I think it's worth at least noting what the env var is

Good point, addressed by https://github.com/scrapinghub/shub/pull/338/commits/32e5a9f321b01f3b5fa23d743b28d0620d8c25a4, I hope it's better now.

Thanks a lot for your input, I agree with everything you've written :raised_hands:

ttilberg:
Great, I think this is a good addition to the documentation. Thanks @vshlapakov. 

Also, as an aside, thanks for taking the time to explain what was going on with the .rst, and even teaching how to compile it as opposed to just being annoyed.

Issue #343, opened 2019-01-15
Cover scrapy-pagestorage in the Custom Images contract documentation


Gallaecio:
Based on the feedback I got in https://github.com/scrapinghub/scrapinghub-entrypoint-scrapy/issues/54, I suspect we should extend https://shub.readthedocs.io/en/stable/custom-images-contract.html#custom-images-contract to cover the need to install `scrapy-pagestorage`.

vshlapakov:
Resolved with https://github.com/scrapinghub/shub/pull/345, thanks for the proposal!

PR #345, opened 2019-02-18
Mention using Scrapy addons in custom image

vshlapakov:
Address https://github.com/scrapinghub/shub/issues/343.

@Gallaecio Could you take a look if it addresses your proposal please?


Gallaecio:
@vshlapakov It does, +1

PR #346, opened 2019-02-21
Add a command to run custom image locally

vshlapakov:
We want to have a separate command to be able to run a custom image locally, for test/debugging purposes. The idea is to mimic (though simplify) SC approach by using a separate wrapper to read crawl process results from FIFO and print it to container stdout. The wrapper is written in Python for simplicity, but could be rewritten with anything more basic (like Bash) in the future.

### DIFF
--- a/shub/image/__init__.py
+++ b/shub/image/__init__.py
@@ -16,6 +16,7 @@ module_deps = [
     "deploy",
     "upload",
     "check",
+    "run",
 ]
 
 for command in module_deps:
--- a/shub/image/run/__init__.py
+++ b/shub/image/run/__init__.py
@@ -1,0 +1,141 @@
+import json
+import stat
+import shlex
+import signal
+import os.path
+from os import stat as os_stat
+from shutil import copyfile
+
+import click
+
+from shub.config import load_shub_config
+from shub.image import utils
+
+
+SHORT_HELP = 'Run custom image locally.'
+HELP = """
+Run a custom Docker image locally.
+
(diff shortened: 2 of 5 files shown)


vshlapakov:
@elacuesta hey man, may I ask you for a quick review in a case I missed something obvious? not sure if you still use the custom image feature, but anyway general code review would be more than enough.

Gallaecio:
When I want to run an image locally, I use Docker directly (`docker build -t <name> . && docker run <name> <command>`).

What advantages does this new command offer over using Docker directly, beyond using a single command instead of two?

vshlapakov:
@Gallaecio That's a very good question. If you take a look into the changes, that's not about running a docker image itself, but mostly about preparing all the environment variables (mainly `SHUB_JOB_DATA`) to run crawl-process in the container similarly to SC platform. No doubt that you could do that manually and construct the env vars by yourself, but it's tricky even for myself. Also, sh-entrypoint-scrapy layer sends all data to SC storage via FIFO file, protocol is described [here](https://doc.scrapinghub.com/scrapy-cloud-write-entrypoint.html#scrapy-cloud-write-entrypoint) if you're interested, and the idea here is to run it in a similar way locally to identify and debug any issues locally - that's especially important for custom images which are not maintained by SC staff.

Gallaecio:
Oh, using the same environment variables as the corresponding Scrapy Cloud project. This is awesome!

vshlapakov:
The feature is quite new, let's move on and address any issues in the following PRs.

Issue #347, opened 2019-03-12
Add poetry support


andrewbaxter:
[poetry](https://github.com/sdispater/poetry) is another popular (at least, frequently mentioned) python package manager, with a lockfile similar to pipenv's.

vshlapakov:
Thanks for creating it, I'm inclined to add its support. I've been watching for the project development for a few months and it looks pretty promising.

andrewbaxter:
@vshlapakov I'd like to try implementing this.

Poetry 1.0.0 (unreleased) has an "export" command that produces a requirements.txt file.  The code to generate it is pretty clean (`Exporter(locker).export('requirements.txt', cwd, use_hashes=False)`) but it's an undocumented api and does include `-e` lines.  Parsing the lockfile and generating the requirements.txt manually is not too difficult (15-20 lines?) and could be done in a way to mirror the Poetry 1.0.0 code.

Any preference on which approach to take?

PR #349, opened 2019-05-10
#347 Add poetry support

andrewbaxter:
I tested with pypi and git packages - the logic as mentioned mirrors code from poetry but I removed the `-e` bits and the hash bits since AFAICT from pipenv those needed to be removed.

### DIFF
--- a/docs/deploying.rst
+++ b/docs/deploying.rst
@@ -82,6 +82,24 @@ A requirements.txt file will be created out of the ``Pipfile`` so like the
 requirements file above, it should not contain packages that are already part
 of the stack.
 
+If you use `Poetry`_ you can specify your ``pyproject.toml``::
+
+    # project_directory/scrapinghub.yml
+
+    projects:
+      default: 12345
+      prod: 33333
+
+    requirements:
+      file: pyproject.toml
+
+A ``poetry.lock`` file must be available, that will be used for determining the
+full requirements.
+
+.. note::
+
+    `Poetry`_ is a tool for dependency management and packaging in Python.
+
 When your dependencies cannot be specified in a requirements file, e.g.
 because they are not publicly available, you can supply them as Python eggs::
 
@@ -104,6 +122,8 @@ build your own Docker image to be used on Scrapy Cloud. See
 .. _requirements file: https://pip.pypa.io/en/stable/user_guide/#requirements-files
 
 .. _pipenv: https://github.com/pypa/pipenv
+
+.. _Poetry: https://poetry.eustace.io/
 
 .. _choose-custom-stack:
 
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -7,6 +7,7 @@ import json
 from six.moves.urllib.parse import urljoin
 
 import click
+import toml
 # Not used in code but needed in runtime, don't remove!
 import setuptools
 _1 = setuptools  # NOQA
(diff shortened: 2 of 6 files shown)


docs/deploying.rst
```diff
@@ -82,6 +82,20 @@ A requirements.txt file will be created out of the ``Pipfile`` so like the
...
+
+    projects:
+      default: 12345
+      prod: 33333
+
+    requirements:
+      file: pyproject.toml
+
+A ``poetry.lock`` file must be available, that will be used for determining the
+full requirements.
```
vshlapakov: As a suggestion: recently I got a request for improving the deploy docs to mention a shortcut to install pipenv  (which was addressed by https://github.com/scrapinghub/shub/commit/f9d31eb9013433e88837d4039853c549de7fef5f), I wonder if it makes sense to do similar thing here, wdyt?

shub/deploy.py
```diff
@@ -187,6 +190,44 @@ def _get_pipfile_requirements():
...
+
+
+def _get_poetry_requirements():
+    try:
+        data = toml.load('poetry.lock')
+    except IOError:
+        raise ShubException('Please make sure the poetry lock file is present')
+    # Adapted from poetry 1.0.0a2 poetry/utils/exporter.py
+    with tempfile.NamedTemporaryFile(
+        dir=os.getcwd(), suffix='.txt', delete=False,
```
vshlapakov: I guess we have to remove the temporary file later somewhere, right? Or maybe return requirements content here instead of a filename and rework `_upload_egg` and skip `open()` call for it (it seems `requests` [allows](https://2.python-requests.org//en/latest/user/quickstart/#post-a-multipart-encoded-file) to pass strings instead of file objects).

docs/deploying.rst
andrewbaxter: Yeah, sure!  I'd assume if they want to use poetry.lock they're already aware of what Poetry is, its official website, etc.  Maybe that request was from the perspective of someone being introduced to a project that uses pipenv but who hasn't previously used the tool?  What about a link to the website, like "\[Poetry](x) is a Python dependency and virtual management tool"

shub/deploy.py
andrewbaxter: Yeah - actually I wasn't sure what was happening in the pipenv code in this case.  Is it creating a requirements.txt file?  Passing the string text itself seems like a nicer way to do it if you don't mind the change scope increasing.

andrewbaxter:
Thanks for the reviews, I'll try to address the comments shortly.

docs/deploying.rst
vshlapakov: > Maybe that request was from the perspective of someone being introduced to a project that uses pipenv but who hasn't previously used the tool? What about a link to the website, like "[Poetry](x) is a Python dependency and virtual management tool"

Yeah, I think so. Sounds good, I'd go with [the official definition](https://poetry.eustace.io/docs/): `is a tool for dependency management and packaging in Python`.

vshlapakov: Also, there's a little dilemma here: from one side, we're trying to keep requirements.txt minimal and avoid shadowing python dependencies already installed in stack (to avoid side effects, speed up deployments etc), so the same rules should be applied to pipfile/poetry (lock) files. From the other side, I assume that pipfile/poetry are used by devs to setup a full environment from scratch, thus it has almost all the deps already installed in stack plus additional ones. And having different sets of files for local development and for SC deployment doesn't sound right. For sure, it's out of the scope of the current PR, so feel free to ignore it for now, but it's something we'll have to think more about in the future.

shub/deploy.py
vshlapakov: >  Is it creating a requirements.txt file? Passing the string text itself seems like a nicer way to do it if you don't mind the change scope increasing.

Yup, it is, and then it returns the temporary requirements filename to the caller which opens it and passes to requests call. It'd be a bit tricky to reliably drop the temporary file later, so passing a string makes sense to me. This is a helper, not a public API, also requirements files are not that large and we already load it to memory while constructing, so I believe it should be safe and cleaner to pass it via string object.

andrewbaxter:
@Gallaecio @vshlapakov Ready for another review I believe.

andrewbaxter:
@vshlapakov squashed now, I believe.

vshlapakov:
Thanks a lot @andrewbaxter! :clap:

PR #351, opened 2019-05-23
Add "no-cache" option on "shub image build" command

dasoran:
enabled to use no-cache option when executing "shub image build" command.

# example

```
$ shub image build --no-cache
```

# details

All I implemented is to pass the no-cache option to the docker client.

### DIFF
--- a/shub/image/build.py
+++ b/shub/image/build.py
@@ -39,13 +39,15 @@ BUILD_SUCCESS_REGEX = re.compile(r'Successfully built ([0-9a-f]+)')
               help="stream build logs to console")
 @click.option("-V", "--version", help="release version")
 @click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
+@click.option("-n", "--no-cache", is_flag=True,
+              help="Do not use cache when building the image")
 @click.option("-f", "--file", "filename", default='Dockerfile',
               help="Name of the Dockerfile (Default is 'PATH/Dockerfile')")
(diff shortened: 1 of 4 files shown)


shub/image/build.py
```diff
@@ -39,13 +39,15 @@
...
 @click.option("-f", "--file", "filename", default='Dockerfile',
               help="Name of the Dockerfile (Default is 'PATH/Dockerfile')")
-def cli(target, debug, verbose, version, skip_tests, filename):
-    build_cmd(target, version, skip_tests, filename=filename)
+def cli(target, debug, verbose, version, skip_tests, no_cache, filename):
+    build_cmd(target, version, skip_tests, no_cache, filename=filename)
 
 
-def build_cmd(target, version, skip_tests, filename='Dockerfile'):
+def build_cmd(target, version, skip_tests, no_cache, filename='Dockerfile'):
```
vshlapakov: There's another `upload` command which uses the `build_cmd` function, could you please update it accordingly? 
- https://github.com/scrapinghub/shub/blob/master/shub/image/upload.py#L48
- https://github.com/scrapinghub/shub/blob/master/tests/image/test_upload.py#L20

dasoran: Thank you for your reviews! @vshlapakov 

I tried to fix `upload` command.

vshlapakov:
@dasoran awesome, thanks!

PR #352, opened 2019-06-04
Feature: add "delete" command to shub

arount:
Hey,

It all started when I was about to send a ticket to Scrapinghub's support to ask for a "remove all" button for pending and finished queue in Scrapinghub's dashboard when I realized I'm developer and I can first code, then gently ask for new feature ;)

This add a `delete` command to `shub` to delete or cancel pending and finished jobs.

```
shub delete -h
Delete or cancel jobs from Scrapy Cloud's project.

A project ID must be passed as argument:

    shub delete 12345

You can select jobs from a specific Spider:

    shub delete 12345 --spider-id=67

Or a specific state ("pending" or "finished" only):

    shub delete 12345 --state=pending

Or both:

    shub delete 12345 --spider-id=67 --state=finished

You can also run it in "force" mode, so it won't ask you permission before flushing jobs.
    shub delete 12345 --force
```

It does not handle running jobs.

Please let me know if anything sound wrong.

### DIFF
--- a/shub/delete.py
+++ b/shub/delete.py
@@ -2,8 +2,6 @@ from __future__ import absolute_import
 import click
 
 from scrapinghub import ScrapinghubClient
-
-from shub.utils import job_resource_iter, get_job
 from shub.config import load_shub_config
 
 HELP = """
@@ -32,6 +30,7 @@ You can also run it in "force" mode, so it won't ask you permission before flush
 SHORT_HELP = "Delete or cancel jobs from Scrapy Cloud"
 
 EXPECTED_STATES = ('pending', 'finished')
+
 
 @click.command(help=HELP, short_help=SHORT_HELP)
 @click.argument('project_id')
--- a/shub/delete.py
+++ b/shub/delete.py
@@ -38,7 +38,6 @@ EXPECTED_STATES = ('pending', 'finished')
 @click.option('--spider-id', required=False)
 @click.option('--state', required=False)
 @click.option('--force', '-f', is_flag=True)
-
 def cli(project_id, spider_id, state, force):
     if state is None:
         state = EXPECTED_STATES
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -6,6 +6,8 @@ dist
 .eggs/
 .tox
 .cache
+env/
```
Gallaecio: I don’t know the policy of the `shub` project, but I think these ignores that are not mandatory of the project are best suited for the developer’s [global git ignore](http://egorsmirnov.me/2015/05/04/global-gitignore-file.html) or [local git ignore](https://stackoverflow.com/a/1753078/939364).

arount: fixed

Gallaecio:
Do you think you can extend the test suite to cover this new feature?

arount:
@Gallaecio ofc, I will try to do this asap.

vshlapakov:
@arount I want to apologize deeply for the lack of communication, but I have to reject this PR in favor of https://github.com/scrapinghub/shub/pull/359, let me explain :confused: 

The problem here is that cancelling a large amount of jobs one-by-one is very slow and inefficient, both from the server and user PoVs, and first of all we had to add the ability to cancel multiple jobs with a single call on server side, then add it to our python client, and finally reuse it here in the library. It doesn't support deletion yet though, something that should be addressed later (it's less critical, as jobs are deleted automatically anyway after a while, according to data retention settings). Another thing is that cancellation and deletion are very different by nature and should be handled separately, I'd like to avoid taking decision what to do just based on the current state of a job.

>  I realized I'm developer and I can first code, then gently ask for new feature ;)

That's cool, thanks for your efforts, you did it well actually, just this specific case is pretty tricky and we had to handle it differently after thorough discussions. Sorry once again, that's my personal failure, I missed the point that the changes are very close to each other, I'll try to be more attentive next time and respond earlier.

vshlapakov:
Superseded by https://github.com/scrapinghub/shub/pull/359.

PR #354, opened 2019-06-27
Fix tests/image/test_run.py::test_consume_from_fifo on non-UTC machines

Gallaecio:
There was a test failure on my machine due to my machine being in a non-UTC timezone.


vshlapakov:
Thanks @Gallaecio ! :+1:

starrify:
This indeed breaks the test for me when using IST (Dublin, Ireland) which has DST (UTC+0/+1) :thinking: 

It breaks in both py27 and py36. One example:
```
E       assert '2017-01-24 1...her message\n' == '2017-01-24 15...her message\n'
E         - 2017-01-24 14:59:01 INFO Some message
E         ?             ^
E         + 2017-01-24 15:59:01 INFO Some message
E         ?             ^
E         - 2017-01-24 14:59:01 WARNING Other message
E         ?             ^
E         + 2017-01-24 15:59:01 WARNING Other message
E         ?             ^

tests/image/test_run.py:140: AssertionError
```

Other quick tests performed:
- Using CST (Shanghai, China) (UTC+8, no DST): Not breaking.
- Using EEST (Kiev, Ukraine) (DST: UTC+2/+3): Not breaking.

IIRC before this change it didn't break for me when using IST.

vshlapakov:
@starrify It seems you're right, the patch doesn't work in all cases. I think it's related with daylight-savings-time: local datetime calculated in the test is 1 hour larger than UTC, while it should be [the same](https://www.worldtimebuddy.com/?qm=1&lid=100,2964574,2988507&h=100&date=2017-1-24&sln=16-16.5) at the specified time. It doesn't affect the main logic of the app in any sense, I believe we need to to simplify the test.
```
In [2]: millis = 1485269941065
In [3]: datetime.datetime.fromtimestamp(millis / 1000.0)
Out[3]: datetime.datetime(2017, 1, 24, 14, 59, 1, 65000)
...
In [11]: local_datetime_string
Out[11]: '2017-01-24 15:59:01'
```

vshlapakov:
@starrify The bug should be fixed by https://github.com/scrapinghub/shub/commit/287138466558cd0255818465a88fd65b71efa9df, thanks for mentioning the issue :+1:

PR #355, opened 2019-06-28
Insecure exception with credential should raise warning

mabelvj:
If we are doing are getting the insecure credentials and we have username or password, an alert will raise.

### DIFF
--- a/shub/image/utils.py
+++ b/shub/image/utils.py
@@ -14,7 +14,7 @@ from shub import config as shub_config
 from shub import utils as shub_utils
 from shub.exceptions import (
     ShubException, NotFoundException, BadConfigException, RemoteErrorException,
-    ShubDeprecationWarning, print_warning
+    ShubDeprecationWarning, print_warning, BadParameterException,
 )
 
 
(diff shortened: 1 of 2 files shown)


shub/image/utils.py
```diff
@@ -157,6 +157,9 @@ def get_credentials(username=None, password=None, insecure=False,
     without 'account' query parameter which breaks login.
     """
     if insecure:
+        if username or password:
+            raise click.BadParameter(
```
vshlapakov: Could you please replace it with [BadParameterException](https://github.com/scrapinghub/shub/blob/master/shub/exceptions.py#L57)? It's nearly the same as original `BadParameter`, but I would prefer to be consistent with the other code.

mabelvj: Sure!

vshlapakov: One thing, the `BadParameterException` lives in `shub.exceptions`, not in `click`

mabelvj: I'm checking right now and in the rest of the function it is used `BarParameter`. For consistency do you refer to other code in shub? Should the lines below 160 too? https://github.com/scrapinghub/shub/blob/6b71fb2ea378e8c659629219617739c9eda286c3/shub/image/utils.py#L168

vshlapakov: Oh, you're right indeed, I missed that. 
> For consistency do you refer to other code in shub?

Yeah, exactly. Could you please replace the other `click.BadParameter` in the function with `BadParameterException` from `shub.exceptions`?

mabelvj:
ok, I will use the merge squash and combine them into a single one

noviluni:
Hi @mabelvj, good job! if you need help with the testing don't hesitate to ask it :)

mabelvj:
Thanks @Gallaecio @noviluni! I've just updated the PR with the tests.

vshlapakov:
Awesome, thanks, merged! And sorry for the delay, I was on vacation :raised_hands:

PR #359, opened 2019-08-06
KUMO-3333: Adding a new command to cancel multiple jobs

rafaelcapucho:
This PR adds a new command called `cancel`, which aims to stop running and awaiting jobs. 
It uses the new `python-scrapinghub`'s `cancel` method.

Usage examples:
  $ shub cancel <project_id> <spider_id>/<job_id> <spider_id>/<job_id> ...
  $ shub cancel <spider_id>/<job_id> <spider_id>/<job_id> ...
  $ shub cancel --force <spider_id>/<job_id> <spider_id>/<job_id> ...

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ ipaddress==1.0.17         # via docker-py
 pyyaml==5.1.1
 requests==2.22.0
 retrying==1.3.3
-scrapinghub==2.1.0
+scrapinghub==2.2.1
 six==1.10.0
 tqdm==4.11.2
 urllib3==1.25.3           # via requests
--- a/shub/cancel.py
+++ b/shub/cancel.py
@@ -1,0 +1,95 @@
+from __future__ import absolute_import
+
+import click
+
+from scrapinghub import ScrapinghubAPIError
+from scrapinghub.client.utils import parse_job_key
+
+from shub.utils import get_scrapinghub_client_from_config
+from shub.config import get_target_conf
+from shub.exceptions import (
+    ShubException,
+    BadParameterException,
+    SubcommandException,
+)
+
+
+HELP = """
+Cancel multiple jobs from Scrapy Cloud.
+
+The cancel command expects the project ID (target) followed by
+the pair containing the spider ID and Job ID:
+
+    shub cancel 12345 1/1 1/2 1/3
+
+If the project ID is not defined it is going to use the default
+project (as defined in scrapinghub.yml):
+
+    shub cancel 1/1 1/2 1/3
+
+The cancel command requires a confirmation that could be skipped
+with the flag --force/-f:
+
+    shub cancel --force 1/1 1/2 1/3
+"""
+
+
+SHORT_HELP = "Cancel multiple jobs from Scrapy Cloud"
+
+
+@click.command(help=HELP, short_help=SHORT_HELP)
+@click.argument("target_or_key")
+@click.argument("keys", nargs=-1)
+@click.option('--force', '-f', is_flag=True,
+              help='It ignores the confirmation prompt')
+def cli(target_or_key, keys, force):
+    # target_or_key contains a target or just another job key
+    if "/" in target_or_key:
+        keys = (target_or_key,) + keys
+        target = "default"
+    else:
+        target = target_or_key
+
+    targetconf = get_target_conf(target)
+    project_id = targetconf.project_id
+    client = get_scrapinghub_client_from_config(targetconf)
+    project = client.get_project(project_id)
+
+    try:
+        job_keys = [validate_job_key(project_id, key) for key in keys]
+    except (BadParameterException, SubcommandException) as err:
+        click.echo('Error during keys validation: %s' % str(err))
+        exit(1)
+
+    if not force:
+        jobs_str = ", ".join([str(job) for job in job_keys])
+        click.confirm(
+            'Do you want to cancel these %s jobs? \n\n%s \n\nconfirm?'
+            % (len(job_keys), jobs_str),
+            abort=True
+        )
+
+    try:
+        output = project.jobs.cancel(
+            keys=[str(job) for job in job_keys]
+        )
+    except (ValueError, ScrapinghubAPIError) as err:
+        raise ShubException(str(err))
+
+    click.echo(output)
+
+
+def validate_job_key(project_id, short_key):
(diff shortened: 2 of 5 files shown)


shub/cancel.py
```diff
@@ -0,0 +1,88 @@
...
+    client = _get_client(targetconf)
+
+    try:
+        project = client.get_project(project_id)
+    except ScrapinghubAPIError as err:
+        raise RemoteErrorException(str(err))
+
+    job_keys = [validate_job_key(project_id, key) for key in keys]
+
+    try:
```
vshlapakov: Cancelling jobs is a risky stuff, let's ask for [confirmation](https://click.palletsprojects.com/en/6.x/prompts/#confirmation-prompts) when we form a list of keys and ready to proceed

shub/cancel.py
```diff
@@ -0,0 +1,88 @@
...
+        return parse_job_key(job_key)
+    except ValueError as err:
+        raise BadParameterException(str(err))
+    except Exception as err:
+        raise SubcommandException(str(err))
+
+
+def _get_client(conf):
```
vshlapakov: Let's move it to shub.utils as`get_scrapinghub_client_from_config`?

shub/cancel.py
```diff
@@ -0,0 +1,88 @@
...
+
+    shub cancel 1/1 1/2 1/3
+"""
+
+
+SHORT_HELP = "Cancel multiple jobs from Scrapy Cloud"
+
+
+@click.command(help=HELP, short_help=SHORT_HELP)
+@click.argument("project_id")
```
vshlapakov: Let's rename it to `target_or_key` to be more clear what's allowed as input

shub/cancel.py
```diff
@@ -0,0 +1,88 @@
...
+    targetconf = get_target_conf(target)
+    project_id = targetconf.project_id
+    client = _get_client(targetconf)
+
+    try:
+        project = client.get_project(project_id)
+    except ScrapinghubAPIError as err:
+        raise RemoteErrorException(str(err))
+
+    job_keys = [validate_job_key(project_id, key) for key in keys]
```
vshlapakov: We should stop early if there's no valid keys for cancellation, probably print a notification about that

shub/cancel.py
```diff
@@ -0,0 +1,88 @@
...
+
+    try:
+        output = project.jobs.cancel(
+            keys=[str(job) for job in job_keys]
+        )
+    except ValueError as err:
+        raise ShubException(str(err))
+
+    click.echo(output)
```
vshlapakov: Wonder if we should wrap the output with human readable message, like "N job(s) were successfully cancelled.", or say that no jobs were cancelled.

tests/test_cancel.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import absolute_import
+
+import unittest
+
+import mock
+
```
vshlapakov: Pls drop the extra line here

requirements.txt
```diff
@@ -15,7 +15,7 @@ ipaddress==1.0.17         # via docker-py
 pyyaml==5.1.1
 requests==2.22.0
 retrying==1.3.3
-scrapinghub==2.1.0
+https://github.com/scrapinghub/python-scrapinghub/archive/7bc63e3.zip
```
vshlapakov: Feel free to use version 2.2.1 from PyPI

shub/cancel.py
rafaelcapucho: Yeah, I thought about that, but in the same time there are other commands like `shub delpoy`.. 
image
But of course we could change it to a human readable message :)

shub/cancel.py
```diff
@@ -0,0 +1,88 @@
...
+    else:
+        target = project_id
+
+    targetconf = get_target_conf(target)
+    project_id = targetconf.project_id
+    client = _get_client(targetconf)
+
+    try:
+        project = client.get_project(project_id)
+    except ScrapinghubAPIError as err:
```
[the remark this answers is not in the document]
rafaelcapucho: You're right, thank you :)

shub/cancel.py
vshlapakov: You're right, we could keep it as-is for now, I think the output should be clear enough :+1:

rafaelcapucho:
Hi @vshlapakov all suggested changes are done, thx! 👍

shub/cancel.py
```diff
@@ -0,0 +1,93 @@
...
+
+    try:
+        job_keys = [validate_job_key(project_id, key) for key in keys]
+    except (BadParameterException, SubcommandException) as err:
+        click.echo('Error during keys validation: %s' % str(err))
+        exit(1)
+
+    if not force:
+        click.confirm(
+            'Do you want to cancel %s jobs?' % len(job_keys), abort=True
```
vshlapakov: Let's also print the job keys itself on the next line: I worry there might be some confusion when using string based targets (or default target), and user will try to cancel wrong jobs without knowing about it.

shub/cancel.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@click.command(help=HELP, short_help=SHORT_HELP)
+@click.argument("target_or_key")
+@click.argument("keys", nargs=-1)
+@click.option('--force', '-f', is_flag=True)
+def cli(target_or_key, keys, force):
+    # Target may be defined or not
+    # When user doesn't specify then the target_or_key
+    # is just another key
```
vshlapakov: That's more a style-level nitpick, but I'd suggest to rephrase and shorten the comment, like:
```
# target_or_key contains a target or just another job key
```

shub/cancel.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+
+SHORT_HELP = "Cancel multiple jobs from Scrapy Cloud"
+
+
+@click.command(help=HELP, short_help=SHORT_HELP)
+@click.argument("target_or_key")
+@click.argument("keys", nargs=-1)
+@click.option('--force', '-f', is_flag=True)
```
vshlapakov: Please add `help` parameter with short explanation what it does

rafaelcapucho: I was able to add the help just to the `@click.option`, because `@click.argument` doesn't accept  :(

vshlapakov: That's fine :+1:

PR #365, opened 2019-08-28
Image push: provide explicit option to re-auth

vshlapakov:
After docker-login step is completed once, it adds a record to its local config with all key data about login, and doesn't try to re-authenticate unless the data is gone. However, we noticed a few cases when explicit re-authentication is required. The PR adds an option to enforce the re-authentication process if needed.

### DIFF
--- a/shub/image/push.py
+++ b/shub/image/push.py
@@ -22,6 +22,8 @@ By default, the tool tries to call the registry in insecure manner,
 otherwise you have to enter your credentials (at least username/password).
 """
 
+LOGIN_ERROR_MSG = 'Please authorize with docker login'
+
 
 @click.command(help=HELP, short_help=SHORT_HELP)
 @click.argument("target", required=False, default="default")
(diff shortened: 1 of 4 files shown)


vshlapakov:
@Gallaecio I'd appreciate if you could take a quick look :raised_hands:

vshlapakov:
I'll add some additional tests a bit later today :ok_hand:

shub/image/push.py
```diff
@@ -39,12 +41,16 @@
 @click.option("--apikey", help="SH apikey to use built-in registry")
 @click.option("--insecure", is_flag=True, help="use insecure registry")
 @click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
+@click.option("-R", "--reauth", is_flag=True,
```
Gallaecio: Why `R` and not `r`?

I’m used to uppercase being used only when lowercase is already used by a different option, which does not seems to be the case. Although I see `S` is in the same situation.

vshlapakov: @Gallaecio  Sorry, I've just realized I missed this one. 

In a general case, lowercase is preferred indeed. However, in a case of internal options which are expected to be used pretty rarely (and can be deprecated and dropped later if the client would work fine w/o it) I'd stay with uppercase, in this way the lowercase alternatives could be used for something more important and viable. Does it make sense? If not really and it's not a matter of preference, I could rename it, no problem.

Gallaecio: Thanks for the explanation, it makes total sense!

vshlapakov:
Thanks for review guys! :+1:

PR #367, opened 2019-09-02
Cleanup intermediate build containers 

hermit-crab:
This change makes `shub image build` behave like `docker build --rm=true .` which has become the default in docker command but was preserved in the client for backward compatibility. I don't know if anyone relies on this behavior so maybe it doesn't make sense but if it does I suppose it's better to shelve this till next major release.


vshlapakov:
Makes sense to me, thanks! :+1:

Issue #366, opened 2019-08-30
Search feature


mirceachira:
It would be awesome to have a search option for shub to look for jobs.

It's often the case that I have to find a broken job starting just from an url and I had to implement a script that cycles through all the jobs and through all the requests of each individual job in a project  to find the one I'm looking for. The search could be configured in many way, see this example:

```
$ my_search_script.py --project_key="<project key>" --spider="amazon" --date_from="last week" --includes="<some url contet>"
> Bingo!  123456/12/123
> Bingo!  987654/32/1
> Done!

```

This could be extended so you could look for a job starting from part of an url, item field, argument etc. You could also search for jobs that have error or critical level logs which would help with debugging.

I thought this might be a good thing to have for shub and would be happy to add it here directly instead of maintaining an internal tool.

mirceachira:
Hey @vshlapakov, I opened a pr for this https://github.com/scrapinghub/shub/pull/370
It has only the search by part of an url feature but I think it's a starting point. Let me know what you think :)

PR #370, opened 2019-09-16
Search feature

mirceachira:
Basic search feature for shub!

There are a million ways to expand this using scrapinghub's api for filtering jobs and requests but I think the core of it (and what is most useful in my case) is to find a project key starting from an url so I thought I'd start with that.

Let me know what you guys think :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ setup(
         'six>=1.7.0',
         'tqdm',
         'toml',
+        'dateparser'
     ],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
--- a/shub/search.py
+++ b/shub/search.py
@@ -1,0 +1,55 @@
+from datetime import datetime
+
+import click
+from dateparser import parse
+from scrapinghub import ScrapinghubClient
+
+
+HELP = """
+Given a project key and part of an url, fetch job ids from Scrapy Cloud.
+
+This is useful when you want to find a job in an efficient way starting from
+an url.
+
+The project key and an url (or part of it). The matching is case sensitive!
+
+    shub search 123456 "B07F3NG1234"
+
+You can provide other parameters to narrow down the search significantly such
+as the spider name and the date interval to search for. Or both! The default
+is to search only the last 6 months.
+    
+    shub search 123456 "B07F3NG1234" --spider="amazon"
+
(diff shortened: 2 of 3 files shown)


shub/search.py
```diff
@@ -0,0 +1,55 @@
...
+SHORT_HELP = "Fetch job ids from Scrapy Cloud based on urls"
+
+
+@click.command(help=HELP, short_help=SHORT_HELP)
+@click.argument('project_key')
+@click.argument('url_content')
+@click.option(
+    '--start_date',
+    default='6 months ago',
+    help='date to start searching from, defaults to 6 months ago'
```
mirceachira: I considered 6 months to be a reasonable default. My main concern is that people will use this without setting spider and start date and it will be slow for them and a lot of requests for scrapinghub cloud but making any of this mandatory might be too restrictive. What do you think?

setup.py
```diff
@@ -39,6 +39,7 @@
         'six>=1.7.0',
         'tqdm',
         'toml',
+        'dateparser'
```
mirceachira: This could be avoided by using a specific format and parsing that, but ofc, this offers more flexibility. Should I remove this?

vshlapakov:
@mirceachira  The feature is really nice and in a great demand, I believe. The core changes also look good in general, good work! However, it requires some additional changes and performance tuning on server side before adding the feature to the official client, so I have to put the PR on hold temporarily while we're discussing what could be done, I'll post an update here or ping your privately when we're ready to proceed. Hopefully, it won't take a long time :crossed_fingers:  

In the meantime, could you please address the Flake8 [complaints](https://travis-ci.org/scrapinghub/shub/jobs/585576345) to fix the build and add some basic tests for the new command? Thanks in advance!

setup.py
vshlapakov: dateparser is a nice and very convenient library, I don't mind to add it, we could use it in the other commands later if needed :+1:

shub/search.py
vshlapakov: I'd say it's too much: we have 120 days of data retention for [the professional plan](https://scrapinghub.com/scrapy-cloud), so it doesn't make sense to have it larger than 4 months anyway. And from the usage perspective, how often do you search for a job that you run a few months ago? I would set the option's default even lower, say a week or two.

starrify:
Implemented job resource filtering in #372

apalala:
fixes #366

PR #373, opened 2019-10-06
Added a "--json" mode (disabled by default) for `shub log`. Useful for handling logs with multi-line contents.

starrify:


starrify:
Hi @vshlapakov thanks for the review! :) I thought of also `--json` initially but decided to make it `--output_json` for being more self-descriptive.
I agree that `--json` may be a better choice since it's simpler, and the PR has been updated accordingly.

vshlapakov:
Merged, thanks! :tada:

PR #384, opened 2020-06-11
Update testing configuration

vshlapakov:
- drop unmaintained Python versions (3.3/3.4)
- start testing with Python 3.7/3.8
- reflect the changes in the docs/setup.py


tox.ini
```diff
@@ -11,7 +11,12 @@ deps =
 
 commands =
     flake8
-    py.test --cov=shub --cov-report= {posargs:shub tests}
+    py.test -p no:warnings --cov=shub --cov-report= {posargs:shub tests}
```
vshlapakov, line 14: @rafaelcapucho  Alright, the CI tests pass now (almost all, a few ones still fail in AppVeyor), but to make it work I had to completely disable warnings: there's some issue happening on python>3.5 that indefinitely retries a test when a warning is yielded, it took me a while before I realized what's happening. While it works, I believe we need to address it before merging the changes. It doesn't happen on py3.5 with exactly the same dependencies, so probably something related to some pytest setting (or any related plugins/dependencies) working differently on py>3.5.

It's reproducible locally  (python 3.6, `pip install -r requirement.txt -r tests/requirements.txt, py.test -v`)
The test I used to reproduce it is this one `test_config.py::LoadShubConfigTest::test_no_global_scrapinghub_yml`, it hangs forever when running locally (and gradually eats all memory available) because of the warnings. If you could take a look and identify/fix the reason, it'd be awesome.

rafaelcapucho: @vshlapakov Hi Vik, The problem is the following, In the version `pytest>=3.1.0` they changed the way they handle warnings:

image

And they said that it might fail when the user's code tries to treat the warning, instead of using the standard, it is what we are doing in `shub/exceptions.py`:

```
def print_warning(msg, category=ShubWarning):
    """Helper to use Python warnings with custom formatter."""

    def custom_showwarning(message, *args, **kwargs):
        # ignore everything except the message
        try:
            sys.stderr.write("WARNING: " + str(message) + '\n')
        # stderr is invalid - this warning just gets lost
        except (IOError, UnicodeError):
            pass

    old_showwarning = warnings.showwarning
    try:
        warnings.showwarning = custom_showwarning
        warnings.warn(msg, category=category)
    finally:
        warnings.formatwarning = old_showwarning
```

Here we are overwriting the `.showwarning` attribute, I did a quick test by allowing the warnings to return again and with:

```
def print_warning(msg, category=ShubWarning):
    warnings.warn(msg, category=category)
```
and the tests started to work again: [qa1wsi9j5siafs4t](https://ci.appveyor.com/project/scrapinghub/shub/builds/33946544/job/qa1wsi9j5siafs4t)

`.showwarning` attribute is used when writing the warning to a file ([Docs](https://docs.python.org/3.8/library/warnings.html#warnings.showwarning)). 

I'm going to think about an approach to sort it.

vshlapakov: Amazing, thanks :100:

shub/exceptions.py
```diff
@@ -107,4 +107,4 @@ def custom_showwarning(message, *args, **kwargs):
         warnings.showwarning = custom_showwarning
         warnings.warn(msg, category=category)
     finally:
-        warnings.formatwarning = old_showwarning
+        warnings.showwarning = old_showwarning
```
rafaelcapucho, line 110: The MemoryError is sorted @vshlapakov  :)

The error was happening because once we're changing `.showwarning` to custom we weren't changing it back, for some reason it was written `.formatwarning` rather than `.showwarning`.

So when pytest were trying to use `showwarning` it was finding something (our custom implementation) and `.formatwarning` wasn't what it is supposed to be anymore :)

vshlapakov: Oh, nice, good catch :+1:

shub/utils.py
```diff
@@ -27,8 +27,11 @@
 # https://github.com/scrapinghub/shub/pull/309#pullrequestreview-113977920
 try:
     from pip import main as pip_main
-except:
-    from pip._internal import main as pip_main
+except:  # noqa
+    try:
+        from pip._internal.main import main as pip_main
+    except ModuleNotFoundError:  # noqa
+        from pip._internal import main as pip_main
```
rafaelcapucho, line 34: @vshlapakov this above changes fixes this error on py38: 
https://ci.appveyor.com/project/scrapinghub/shub/builds/33969869/job/veplrw43tgu513o4

Reference:
https://stackoverflow.com/questions/43398961/pip-module-has-no-attribute-main/49853826#49853826 

Now the only failing test are this on py27: 
https://ci.appveyor.com/project/scrapinghub/shub/build/job/3wp6gr9kpxgoxji5

We're almost there :)

vshlapakov: Cool :+1:  About the last failed test, it seems related to the recently merged https://github.com/pallets/click/pull/1591, should be fixed in the next version of click. Not sure what to do :thinking: I see it happens on Python2.7 / Windows only, so the scope is probably limited, but still it seems to be a breaking change (triggered by the first invocation of click.echo). Also, `click=8.0` is going to drop support for Python 2 anyway.

vshlapakov: Python 2 already hit the EoL, so I believe it should be fine to officially stop supporting it in the next minor version (`2.11`) - remove CI tests, clean setup.py etc, and do the rest of the cleanup in related PRs.

rafaelcapucho: @vshlapakov Hey Vik, my suggestion is to stick with `click==7.0` at least until we have their hotfix released, I did test with `v7.0` and it isn't throwing any error, with this we could conclude our upgrade and release the newer shub, wdyt? cya

vshlapakov: @rafaelcapucho I see, this is a very good point, awesome that you found it :+1: Yeah, the breaking change was introduced in 7.1 version by https://github.com/pallets/click/issues/1065 to fix specific click behavior (redirecting output to NUL on windows), let's stick with `7.0` - at least the basic functionality seems to work - and move away from Python 2 later.

vshlapakov:
@rafaelcapucho Awesome, thanks a lot for help! :clap:

PR #385, opened 2020-06-30
It tries to fix pip errors on newer version

rafaelcapucho:
Creating a PR to force the build in appveyor.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,13 +11,6 @@ with io.open(os.path.join(here, 'shub', '__init__.py'),
     exec(f.read(), about)
 
 
-extra_requirements = []
-try:
-    import pip  # NOQA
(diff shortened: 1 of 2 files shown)


vshlapakov:
@rafaelcapucho  To fix the AppVeyour build, I believe you should take a look into the pip-related logic in `shub/utils.py` and make it work with latest pip versions too - note it's 18.1 in Travis and 19.3.1 in AppVeyor, so the shub code doesn't support some changes made in the recent pip. Version 18.1 got released in the end of 2018, I believe it should be fine to support only pip>=19 in 2020.

vshlapakov:
@rafaelcapucho Awesome, AppVeyor builds are fixed now, thanks :+1:

PR #388, opened 2020-09-10
allow to pass build arguments to docker client

kalessin:
Needed for passing things like private credentials without hard coding them in Dockerfile.

### DIFF
--- a/shub/image/build.py
+++ b/shub/image/build.py
@@ -41,19 +41,22 @@ BUILD_SUCCESS_REGEX = re.compile(r'Successfully built ([0-9a-f]+)')
 @click.option("-S", "--skip-tests", help="skip testing image", is_flag=True)
 @click.option("-n", "--no-cache", is_flag=True,
               help="Do not use cache when building the image")
+@click.option("-b", "--build-arg", multiple=True,
+              help="Allow to pass build arguments to docker client.")
 @click.option("-f", "--file", "filename", default='Dockerfile',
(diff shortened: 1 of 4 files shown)


shub/image/build.py
```diff
@@ -41,19 +41,22 @@
...
 
-def build_cmd(target, version, skip_tests, no_cache, filename='Dockerfile'):
+def build_cmd(target, version, skip_tests, no_cache, build_arg, filename='Dockerfile'):
     config = load_shub_config()
     create_scrapinghub_yml_wizard(config, target=target, image=True)
     client = utils.get_docker_client()
     project_dir = utils.get_project_dir()
     image = config.get_image(target)
     image_name = utils.format_image_name(image, version)
+    build_args = dict(a.split('=') for a in build_arg)
```
vshlapakov, line 59: In theory, there might be build args with values having "=" inside, let's split by the first character only?
```suggestion
    build_args = dict(a.split('=', 1) for a in build_arg)
```

vshlapakov (commented):
LGTM, added a minor suggestion :+1:

shub/image/build.py
kalessin: you are right. I will fix.

kalessin: Done.

Gallaecio (approved):
Nice!

vshlapakov (approved):

vshlapakov:
Thanks @kalessin :+1:

Issue #389, opened 2020-12-07
shub deploy does not run on windows


shyim:
```
c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\setuptools\distutils_patch.py:25: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
  warnings.warn(
Building images.scrapinghub.com/project/485868:1607361205.
Traceback (most recent call last):
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\shyim\AppData\Local\Programs\Python\Python39\Scripts\shub.exe\__main__.py", line 7, in <module>
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\shub\deploy.py", line 78, in cli
    upload_cmd(target, version)
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\shub\image\upload.py", line 54, in upload_cmd
    build.build_cmd(target, version, skip_tests, no_cache, build_arg, filename=filename)
  File "c:\users\shyim\appdata\local\programs\python\python39\lib\site-packages\shub\image\build.py", line 80, in build_cmd
    build_progress = build_progress_cls(events)
[...]

PR #390, opened 2021-01-04
Upgrade tqdm

slygent:
tqdm 4.11.2 is incompatible with python 3.9 due to sys.setcheckinterval being removed, see #389

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -6,7 +6,7 @@ six
 toml
 
 click==7.0
-tqdm==4.11.2
+tqdm
 scrapinghub>=2.3.1
 
 # address known vulnerabilities


rafaelcapucho:
Related: https://github.com/scrapinghub/shub/pull/383#pullrequestreview-436386760

basically newer versions of `tqdm` generates a little bit different progress bars, which fails in our unittests, that is why it was pinned in the first place (as updateing tqdm's testcases wasn't the main target of the task).

we can upgrade `tqdm` but it would also require a fix/rework in the test cases, or remove the tqdm's test cases, I'm not so sure why we were testing the progress bar ourselfs.

slygent:
I don't see any tests on travis for python 3.9 :) If you plan to support shub on python 3.9 (first released in October) then something will have to done about this old version of tqdm.

rafaelcapucho:
hi @slygent, I'm working on this now, It is gonna be fix soon, thx for the reporting

rafaelcapucho:
I'm closing this PR, it was solved in a separated PR: https://github.com/scrapinghub/shub/pull/391

It should be released soon in a new version of shub, thanks for the reporting @slygent !

PR #391, opened 2021-01-07
It adds support to Python 3.9

rafaelcapucho:


tests/image/test_deploy.py
```diff
@@ -174,6 +174,7 @@ def test_progress_bar_logic():
         'Progress:  65%|██████▌   | 65/100'
         'Progress:  75%|███████▌  | 75/100'
         'Progress: 100%|██████████| 100/100'
+        'Progress: 100%|██████████| 100/100'
```
rafaelcapucho, line 177: The duplication is required because newer version of `tqdm` reprints the progress bar when closing:

image

rafaelcapucho:
I'm currently trying to solve this, pytest says that `a` and `b` are different, but I'm not finding the difference, they have the same size, type, I removed all the `\n` and `\r` using `re.sub(r'\r', '', a)`.

I hope to find a solution soon this Monday

image

rafaelcapucho (commented):
Progress Bar and Python 3.9 is working 😁

image

ruairif (approved):
Just gave it a quick review but I don't see any major issues

vshlapakov (approved):
Looks great, thanks :tada:

PR #407, opened 2021-09-08
[dependencies] update 3rd party dependencies

### LINKED ISSUE #404: relax click dependency to make it possible to satisfy other project dependencies
Currently `click` is pinned at version 7.0, which makes satisfying all dependency constraints for a project that also has shub installed difficult, especially given that click 7.0 was released in September 2018, with the latest version being 8.0.1.

Is there a reason for pinning click? If not, could this be relaxed to make it easier to use with other packages in the same virtual environment?

If, on the other the shub package is not intended to be an installed library and rather just an executable, perhaps the documentation need to make this clear, and suggest installing the package with a tool like pipx, which creates a standalone virtual environment, rather than with pip into the current environment.

apalala:
Remove pins to very old versions that produce conflicts with other tools used alongside Scrapy (fixes #404).

Remove `requirements.in` and rely on `setyp.py.install_requires`.

Remove `pip` from dependencies as mentioning it is deemed unsafe.

Upgrade all libraries through `pip-compile --upgrade`.

Hopefully solve security warnings reported by [dependabot](https://github.com/scrapinghub/shub/security/dependabot)


.travis.yml
```diff
@@ -48,10 +48,11 @@ matrix:
...
 
-branches:
-  only:
-    - master
-    - /^v\d+\.\d+\.\d+[\w\-]*$/
+#branches:
+#  only:
+#    - master
+#    - /^v\d+\.\d+\.\d+[\w\-]*$/
+#    - update_dependencies
```
Gallaecio, line 55: Debug leftover?

setup.py
```diff
@@ -31,7 +31,6 @@
     install_requires=[
         'click==7.0',
```
Gallaecio, line 32: I believe this is the `click` pinning that matters, the one affecting `pip install shub`.

shub/utils.py
```diff
@@ -830,6 +830,6 @@ def create_scrapinghub_yml_wizard(conf, target='default', image=None):
     if image or (image is None and _detect_custom_image_project()):
         repository = click.prompt(
             "Image repository (leave empty to use Scrapinghub's repository)",
-            default=True, show_default=False)
+            default=repository, show_default=False)
```
Gallaecio, line 833: Could you elaborate why we are changing `True` to `None` here, how this affects behavior? Is this something we need to do before rebranding? (I’m just wary of behavior changes breaking code in corner cases)

setup.py
apalala: `click` cannot be `>=8.0` for now because Python 2.7 is not supported.

shub/utils.py
apalala: This was an incorrect change. `shub` actually depends on `default=True` as a default.

.travis.yml
apalala: cleared

setup.py
Gallaecio: Still, can it be higher? e.g. `click >=7, <8`?

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.13.0
+current_version = 2.14.0
```
Gallaecio, line 2: I don’t think we should use `bumpversion` as part of a pull request. Current (best?) practice is to do that as part of the release process.

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+        python-version: 3.8
+        python-version: 3.9
```
Gallaecio, line 24: Why not test the same as Travis CI?

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest tox
+        pip install -r tests/requirements.txt
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
Gallaecio, line 35: Why not use `tox` instead of using `flake8` directly, and keep this within the `tox` configuration so that running `tox -e flake8` has the same effect, and users can run that locally?

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: python
-python: 3.6
+python: 3.9
```
Gallaecio, line 2: Why keep both Travis CI and GitHub Actions?

setup.py
```diff
@@ -29,15 +29,14 @@
     include_package_data=True,
     zip_safe=False,
     install_requires=[
-        'click==7.0',
+        'click>=7.1.2',
```
Gallaecio, line 32: Why not support `7.0`, do we use something that is only available in 7.1.2 and higher?

setup.py
```diff
@@ -46,9 +45,6 @@
         'Natural Language :: English',
         'License :: OSI Approved :: BSD License',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
```
Gallaecio, line 51: I’m not sure about dropping Python support. I think it may be safer to do this _after_ the rebranding.

tests/test_utils.py
```diff
@@ -505,6 +505,7 @@ def get_project_dir():
                 os.path.join(basepath, 'a', 'b'))
 
 
+@unittest.skip('broken by changes in `click.invoke(input=)`')
```
Gallaecio, line 508: Can’t it be fixed instead?

setup.py
apalala: This was already changed to unbound `click` after skipping the failed tests.

.bumpversion.cfg
apalala: I'll revert the `bumpversion`.

At any rate, we should be releasing a `3.0.0` because of the loss of compatibility with 2.7

.github/workflows/python-app.yml
apalala: I will remove 2.7, 3.5, and 3.6 from TravisCI.

Yet using Github Actions is experimental, and it's not working. I'll probably revert it.

.travis.yml
apalala: Github Actions are not yet working properly.

setup.py
apalala: Changed to `click`, which means that `>=8.0` will be used.

tests/test_utils.py
apalala: I did many tests, and even went to look at the implementation of `click.testing.invoke(input=)` in 7.0.

I'm sure that the bug is in `click`, because `>=8.0` tests don't use the `input=` feature any more.

The skipped tests address interactive prompting by `shub`. We can skip them safely for a release that upgrades dependencies.

setup.py
apalala: We must at least drop 2.7 to address outdated dependencies. 

If we drop 2.7, we should take the chance to drop 3.5 and 3.6, since no projects should be using them.

If there are projects with 2.7, 3.5, or 3.6, they should be able to use the current version of `shub` as is.

.github/workflows/python-app.yml
apalala: Github Actions were an experiment. We don't need them now, so I'll remove them.

hugovk:
Tip: add `python_requires=">=3.6",` to `setup.py` to help pip know which library version to install for end users.

.github/workflows/python-app.yml
Gallaecio: Are we delaying the next release until 3.6 goes end of life on 23 Dec?

Gallaecio: Regarding GitHub Actions, it’s the only way forward, we cannot use Travis CI anymore.

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.13.0
+current_version = 3.0.0
```
Gallaecio, line 2: https://github.com/scrapinghub/shub/pull/407#discussion_r705016636

.travis.yml
Gallaecio: https://github.com/scrapinghub/shub/pull/407#discussion_r716021976

requirements.txt
```diff
@@ -1,43 +1,43 @@
...
+    #   shub (setup.py)
+toml==0.10.2
+    # via shub (setup.py)
+tqdm==4.62.3
+    # via shub (setup.py)
+urllib3==1.26.7
     # via requests
-websocket-client==0.57.0
+websocket-client==1.2.1
     # via docker
```
Gallaecio, line 43: Do we need this file at all? Given this is a library, `setup.py` should do it.

setup.py
```diff
@@ -29,15 +29,14 @@
     include_package_data=True,
     zip_safe=False,
     install_requires=[
-        'click==7.0',
+        'click',
```
Gallaecio, line 32: Unless you are sure our code works with any version of click, we should keep the old pinned version as a lower limit.

```suggestion
        'click>=7',
```

setup.py
```diff
@@ -29,15 +29,14 @@
...
         'docker',
-        'pip',
-        'PyYAML',
+        'pyyaml',
         'retrying',
         'requests',
         'scrapinghub>=2.3.1',
         'six>=1.7.0',
-        'tqdm==4.55.1',
+        'tqdm',
```
Gallaecio, line 39: We should probably keep some limit, assuming we do not support just any version.

setup.py
Gallaecio: Given the drop of 3.6, we should hold this pull request until Dec 23.

tests/requirements.txt
```diff
@@ -1,31 +1,67 @@
...
+    # via
+    #   pytest
+    #   pytest-cov
+virtualenv==20.7.2
+    # via pipenv
+virtualenv-clone==0.5.7
+    # via pipenv
 
 # The following packages are considered to be unsafe in a requirements file:
 # pip
```
Gallaecio, line 67: It probably makes sense to remove this file in favor of `tox.ini`.

tests/test_utils.py
Gallaecio, line 508: Then we should either address this in code or limit the upper version of `click` to a version we support in `setup.py`.

PR #408, opened 2021-09-08
[rebranding] scrapinghub.com -> zyte.com

apalala:
Switch `scrapinghub.com` for `zyte.com`:

* pending https://images.scrapinghub.com because https://images.zyte.com is not configured
* http://dash.scrapinghub.com/api/settings/get.json mentioned in docs, yet seems invalid

`scrapinghub.yml` is mentioned _all over the place_, so this PR will not address that name.


Gallaecio:
I would personally delay this change until after the rebranding of the project itself.

apalala:
> I would personally delay this change until after the rebranding of the project itself.

I agree is not the right place or time for the rebranding.

PR #413, opened 2021-12-02
Run CI on GitHub Actions

elacuesta:
The idea is to stick to the title and avoid overlapping with #407 as much as possible, but a few things might be duplicated (Flake8 adjustments and Python classifiers, for instance). 

- [x] Flake8 job (small adjustments)
- [x] Tests (Python 3.6 to 3.9; Ubuntu, macOS, Windows)
- [x] Readme badges
- [x] Removed Python 2.7 and 3.5 classifiers
- [x] PyPI publishing job
- [x] Add PyPI token
- [x] Attach artifacts to (draft) release (binaries created with the "freeze" tox environment)

Commit history is not very clean, it should be squash-merged.


tests/test_deploy_egg.py
```diff
@@ -56,6 +56,7 @@ def test_can_clone_a_git_repo_and_deploy_the_egg(self):
 
         self.assertTrue('master' in data['version'])
 
+    @unittest.skip('flaky')
```
elacuesta, line 59: This test works when executed on its own, but fails when executed as part of the whole test suite. While it would be great to know exactly what causes this, at this point I think it's more important to restore CI to a functioning state.

Gallaecio (commented):
The code looks good to me, my only concern is one I expressed in https://github.com/scrapinghub/shub/pull/407:

> I’m not sure about dropping Python support. I think it may be safer to do this _after_ the rebranding.

It may not be a valid concern anymore, though.

Does Scrapy Cloud still support those removed Python versions? If so, we could still go ahead with this change, but we should probably include the python version requirement into `setup.py`, so that users with a modern `pip` get a version of `shub` compatible with Python 2, and also update documentation resources about installing shub to take into account the possibility that users have an old `pip` and hence they should manually pin the `shub` version in the `pip install` command.

ejulio (approved):

wRAR (approved):

PR #415, opened 2022-03-10
Python 3.10 support

### LINKED ISSUE #412: Add support for Python 3.10
At the moment users will get:
`ImportError: cannot import name 'MutableMapping' from 'collections' (C:\Python310\lib\collections\__init__.py)`

This seems to be because of [this dependency](https://github.com/scrapinghub/python-hubstorage/blob/master/hubstorage/resourcetype.py) (line 4) which needs to be:
`from collections.abc import MutableMapping`

elacuesta:
Closes #412


tox.ini
```diff
@@ -19,6 +19,7 @@ deps =
     pyinstaller==4.0
     pytest
     packaging==20.4
+    setuptools==59.8.0  # https://github.com/pypa/setuptools/issues/3089
```
Gallaecio, line 22: Upstream seems to be looking for a reproducible example for this bug. Any chance you can work one out? I fear they may close the issue without addressing it otherwise.

Gallaecio (approved):

PR #416, opened 2022-03-18
Allow click versions other than 7.0

### LINKED ISSUE #404: relax click dependency to make it possible to satisfy other project dependencies

Gallaecio:
Fixes #404

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -24,6 +24,7 @@ import click
 import pip
 import requests
 import yaml
+from click import ParamType
 
 # https://github.com/scrapinghub/shub/pull/309#pullrequestreview-113977920
 try:
@@ -785,6 +786,20 @@ def _update_conf_file(filename, target, project, repository):
         click.echo("Saved to %s." % filename)
 
 
+class _BooleanOrStringParamType(ParamType):
+
+    def convert(self, value, param, ctx):
+        if isinstance(value, bool):
+            return value
+        value = str(value)
+        lowercase_value = value.lower()
+        if lowercase_value == 'true':
+            return True
+        if lowercase_value == 'false':
+            return False
+        return value
+
+
 def create_scrapinghub_yml_wizard(conf, target='default', image=None):
     """
     Ask user for project ID, ensure they have access to that project, and save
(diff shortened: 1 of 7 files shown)


shub/utils.py
```diff
@@ -830,6 +845,6 @@ def create_scrapinghub_yml_wizard(conf, target='default', image=None):
     if image or (image is None and _detect_custom_image_project()):
         repository = click.prompt(
             "Image repository (leave empty to use Scrapinghub's repository)",
-            default=True, show_default=False)
+            default=True, show_default=False, type=_BooleanOrStringParamType())
```
Gallaecio, line 848: This was failing since click 8.0.1, specifically since https://github.com/pallets/click/commit/ad52b19c84ab47cf6ec4df25b4e89ac939c3b253, which due to `default=True` was converting string inputs to `True` as well.

tox.ini
```diff
@@ -6,10 +6,14 @@ basepython = python3
...
-    -r{toxinidir}/tests/requirements.txt
+    -r tests/requirements.txt
 commands =
-    pytest --verbose --cov=shub --cov-report=term-missing --cov-report=html --cov-report=xml {posargs:shub tests}
+    pytest --cov=shub --cov-report=term-missing --cov-report=html --cov-report=xml {posargs:shub tests}
+
+[testenv:min]
+deps =
+    {[testenv]deps}
+    -r requirements.txt
```
Gallaecio, line 16: A separate Tox environment to test those “minimum” versions of dependencies, while keeping the rest of the Tox environments testing the latest versions of dependencies.

elacuesta (approved):
:rocket:

shub/utils.py
```diff
@@ -785,6 +786,12 @@ def _update_conf_file(filename, target, project, repository):
         click.echo("Saved to %s." % filename)
 
 
+class _AnyParamType(ParamType):
```
Gallaecio, line 789:
```suggestion
class _AnyParamType(ParamType):
    name = "any"
```

Issue #417, opened 2022-04-11
pipenv.utils.convert_deps_to_pip removed


wRAR:
`shub.deploy._get_pipfile_requirements()` uses `pipenv.utils.convert_deps_to_pip()` which was moved to `pipenv.utils.dependencies.convert_deps_to_pip()` in the 2022.4.8 version and is also considered internal. See https://github.com/pypa/pipenv/issues/5040

elacuesta:
#418 released with `shub==2.14.1` as a workaround, let's keep this issue open as a task for replacing these functions with pipenv's public API.

alepuccetti:
I am using `shub==2.14.2` and when trying to deploy with a Pipfile I get this error:
```
File "~/.local/share/virtualenvs/my-project-4mAwm4Xi/lib/python3.9/site-packages/shub/deploy.py", line 193, in _get_pipfile_requirements
    return open(_add_sources(convert_deps_to_pip(deps), _sources=sources.encode(), tmpdir=tmpdir), 'rb')
  File "~/.local/share/virtualenvs/my-project-4mAwm4Xi/lib/python3.9/site-packages/shub/deploy.py", line 199, in _add_sources
    with open(_reqs_file, 'rb') as f:
TypeError: expected str, bytes or os.PathLike object, not list
```

I did some digging and it seems that `convert_deps_to_pip(deps)` returns a list into `_reqs_file` (fisrt argument of `_add_sources`) and this makes fail the `open` as in the error stack I provided above.
The error happen before reaching the cloud in the preparation stage. The only workaround I found was to create a `requirements.txt` file and run `shub deploy` using that file.

elacuesta:
Please try again with [`v2.14.3`](https://github.com/scrapinghub/shub/releases/tag/v2.14.3)

PR #418, opened 2022-04-12
Import pipenv utils from new location

elacuesta:
Workaround for #417. Not a complete fix, we should look into replacing these functions, maybe with the `requirements` or `lock` commands.


rafaelcapucho (approved):
Thanks!

PR #422, opened 2022-10-28
Fix deps conversion

alexandresgf:
Since Pipenv v2022.8.30 the function `convert_deps_to_pip` returns a list of strings containing all the dependencies found in the Pipfile converted to Pip format. So, I added a try/exception statement to catch the `TypeError` exception, when the previous code tries to open the temporary file, and write the dependencies to the requirements file.

### DIFF
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -11,11 +11,11 @@ import click
 # Not used in code but needed in runtime, don't remove!
 import setuptools
 import setuptools.msvc  # noqa
+
 import toml
 from six.moves.urllib.parse import urljoin
 
-from shub.config import (SH_IMAGES_REGISTRY, list_targets_callback,
-                         load_shub_config)
+from shub.config import (SH_IMAGES_REGISTRY, list_targets_callback, load_shub_config)
 from shub.exceptions import (BadParameterException, NotFoundException,
                              ShubException)
 from shub.image.upload import upload_cmd
@@ -170,8 +170,7 @@ def _get_pipfile_requirements(tmpdir=None):
         from pipenv.utils.indexes import prepare_pip_source_args
     except ImportError:
         try:
-            from pipenv.utils import (convert_deps_to_pip,
-                                      prepare_pip_source_args)
+            from pipenv.utils import (convert_deps_to_pip, prepare_pip_source_args)
         except ImportError:
             raise ImportError('You need pipenv installed to deploy with Pipfile')
     try:
--- a/shub/deploy.py
+++ b/shub/deploy.py
@@ -196,7 +196,7 @@ def _get_pipfile_requirements(tmpdir=None):
 def _add_sources(_reqs_file, _sources, tmpdir=None):
     tmp = tempfile.NamedTemporaryFile(delete=False, suffix="-requirements.txt", dir=tmpdir)
     tmp.write(_sources + b'\n')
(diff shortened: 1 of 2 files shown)


elacuesta:
Hi, thanks for the fix. I have a few comments:
* Please split the code paths based on whether the received argument is a list - I don't think catching `TypeError` makes this distinction clear.
* Joining a variable called `_reqs_file` is confusing to me, please rename the parameter to show that it can be either a string or a list. It would be even better if you could also add type hints.
* Could this be tested? I guess it would be enough to add an environment with a different pipenv version, without actually modifying any test cases.

alexandresgf:
> * Could this be tested? I guess it would be enough to add an environment with a different pipenv version, without actually modifying any test cases.

I can mock-up the behavior of both `convert_deps_to_pip` and write the tests. Is it sounds good?

elacuesta:
Sounds good to me, thanks!

shub/deploy.py
```diff
@@ -169,7 +170,8 @@ def _get_pipfile_requirements(tmpdir=None):
         from pipenv.utils.indexes import prepare_pip_source_args
     except ImportError:
         try:
-            from pipenv.utils import convert_deps_to_pip, prepare_pip_source_args
+            from pipenv.utils import (convert_deps_to_pip,
+                                      prepare_pip_source_args)
```
elacuesta, line 174: Please revert this change, we don't have such line length restriction in place.

shub/deploy.py
```diff
@@ -1,25 +1,26 @@
...
 # Not used in code but needed in runtime, don't remove!
 import setuptools
 import setuptools.msvc  # noqa
+import toml
+from six.moves.urllib.parse import urljoin
 
-from shub.config import (list_targets_callback, load_shub_config,
-                         SH_IMAGES_REGISTRY)
+from shub.config import (SH_IMAGES_REGISTRY, list_targets_callback,
+                         load_shub_config)
```
elacuesta, line 18: Please put this in one line

shub/deploy.py
```diff
@@ -1,25 +1,26 @@
...
-import json
-from six.moves.urllib.parse import urljoin
+from typing import AnyStr, Optional, Union
 
 import click
-import toml
 # Not used in code but needed in runtime, don't remove!
 import setuptools
 import setuptools.msvc  # noqa
+import toml
```
elacuesta, line 14: I appreciate the attempt to clean up the imports, however let's keep the whiteline after the setuptools imports - the comment above only applies to them.

elacuesta (commented):
Great job, thanks. I left a minor comments.

shub/deploy.py
alexandresgf: Sorry, it is an addiction that I have to format the code after making any changes. I have already returned it to the previous state.

shub/deploy.py
alexandresgf: Sorry, it is an addiction that I have to format the code after making any changes. I have already returned it to the previous state.

shub/deploy.py
alexandresgf: Sorry, it is an addiction that I have to format the code after making any changes. I have already returned it to the previous state.

elacuesta:
Thanks @alexandresgf :partying_face:

PR #434, opened 2023-11-30
add support for python 3.12

branchvincent:
Adds support for Python 3.12. Main changes:

- `NamedTemporaryFile` no longer has a `delete` attribute, presumably this wasn't part of the public API: https://github.com/python/cpython/commit/743453a5541d3a6f19046459d22a7991ecb88223#diff-6553a99f3ae04c9fc9c2349ac27037bfe3b274d5ca54d5800d9c5b5f11e29d21L470 (refactored to not rely on it)
- Deprecated assertion helpers were [removed](https://docs.python.org/3/whatsnew/3.12.html#id3) (simple renames)
- `distutils` was [removed](https://docs.python.org/3/whatsnew/3.12.html#distutils) (replaced version classes with equivalent one from `packaging`)
- `pkg_resources` is [no longer provided by default](https://docs.python.org/3/whatsnew/3.12.html#ensurepip) (was also deprecated, so replaced it with `importlib.metadata`)
- Deprecated `configparser.SafeConfigParser` was [removed](https://docs.python.org/3/whatsnew/3.12.html#configparser) in favor of `ConfigParser`

### DIFF
--- a/shub/utils.py
+++ b/shub/utils.py
@@ -11,9 +11,9 @@ import re
 import time
 
 from collections import deque
-from six.moves.configparser import SafeConfigParser
-from distutils.spawn import find_executable
-from distutils.version import LooseVersion, StrictVersion
+from configparser import ConfigParser
+from shutil import which
+from packaging.version import Version
 from glob import glob
 from importlib import import_module
 from tempfile import NamedTemporaryFile, TemporaryFile
@@ -143,17 +143,21 @@ def _check_deploy_files_size(files):
 
 def write_and_echo_logs(keep_log, last_logs, rsp, verbose):
     """It will write logs to temporal file and echo if verbose is True."""
-    with NamedTemporaryFile(prefix='shub_deploy_', suffix='.log',
-                            delete=(not keep_log)) as log_file:
-        for line in rsp.iter_lines():
-            if verbose:
-                click.echo(line)
-            last_logs.append(line)
-            log_file.write(line + b'\n')
+    log_contents = b""
+    for line in rsp.iter_lines():
+        if verbose:
+            click.echo(line)
+        last_logs.append(line)
+        log_contents += line + b'\n'
+    deployed = _is_deploy_successful(last_logs)
+    if not deployed:
+        keep_log = True
+    echo_short_log_if_deployed(deployed, last_logs, verbose=verbose)
 
-        deployed = _is_deploy_successful(last_logs)
-        echo_short_log_if_deployed(deployed, last_logs, log_file, verbose)
-        if not log_file.delete:
+    with NamedTemporaryFile(prefix='shub_deploy_', suffix='.log',
+                            delete=not keep_log) as log_file:
+        log_file.write(log_contents)
+        if keep_log:
             click.echo("Deploy log location: %s" % log_file.name)
         if not deployed:
             try:
@@ -163,12 +167,11 @@ def write_and_echo_logs(keep_log, last_logs, rsp, verbose):
             raise RemoteErrorException("Deploy failed: {}".format(last_log))
 
 
-def echo_short_log_if_deployed(deployed, last_logs, log_file, verbose):
+def echo_short_log_if_deployed(deployed, last_logs, log_file=None, verbose=False):
     if deployed:
         if not verbose:
             click.echo(last_logs[-1])
     else:
-        log_file.delete = False
         if not verbose:
             click.echo("Deploy log last %s lines:" % len(last_logs))
             for line in last_logs:
@@ -212,7 +215,7 @@ def patch_sys_executable():
 
 
 def find_exe(exe_name):
-    exe = find_executable(exe_name)
+    exe = which(exe_name)
     if not exe:
         raise NotFoundException("Please install {}".format(exe_name))
     return exe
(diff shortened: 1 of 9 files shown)


shub/utils.py
```diff
@@ -143,17 +143,19 @@ def _check_deploy_files_size(files):
...
+    deployed = _is_deploy_successful(last_logs)
+    keep_log = echo_short_log_if_deployed(deployed, last_logs, keep_log, verbose)
 
-        deployed = _is_deploy_successful(last_logs)
-        echo_short_log_if_deployed(deployed, last_logs, log_file, verbose)
-        if not log_file.delete:
+    with NamedTemporaryFile(prefix='shub_deploy_', suffix='.log',
+                            delete=not keep_log) as log_file:
+        log_file.write(log_contents)
+        if keep_log:
```
Gallaecio, line 158: How does this change relate to 3.12?

branchvincent: > `NamedTemporaryFile` no longer has a `delete` attribute

Here `write_and_echo_logs` tries to get that attribute:

https://github.com/scrapinghub/shub/blob/0f3087f32b81c4ac6a966bb23f0c2414c349eff6/shub/utils.py#L156

On 3.12, that throws `AttributeError: '_io.BufferedRandom' object has no attribute 'delete'`. Also `echo_short_log_if_deployed` sets `delete` when the deploy failed which no longer makes sense (in fact, this didn't have the intended effect of keeping the file since `__exit__` didn't respect it, only [`_closer.delete`](https://github.com/python/cpython/blob/3.11/Lib/tempfile.py#L509)):

https://github.com/scrapinghub/shub/blob/0f3087f32b81c4ac6a966bb23f0c2414c349eff6/shub/utils.py#L171

So instead, this change just waits to construct the temp file until the final value for `delete` is determined

shub/utils.py
```diff
@@ -163,16 +165,17 @@ def write_and_echo_logs(keep_log, last_logs, rsp, verbose):
             raise RemoteErrorException("Deploy failed: {}".format(last_log))
 
 
-def echo_short_log_if_deployed(deployed, last_logs, log_file, verbose):
+def echo_short_log_if_deployed(deployed, last_logs, keep_log, verbose):
```
Gallaecio, line 168: This is a backward-incompatible change. Can we restore backward compatibility? (i.e. add a new parameter, maybe log a deprecation warning if `log_file` is not `None` or similar)

Gallaecio:
cc @wRAR

tests/test_utils.py
```diff
@@ -66,15 +66,15 @@ def test_run_cmd_captures_stderr(self):
...
+        with self.assertRaisesRegex(SubcommandException, r'STDERR[\s-]+Hello'):
             cmd[-1] += '; sys.exit(99)'
             utils.run_cmd(cmd)
 
     def test_pwd_git_version_without_git(self):
         # Change into test dir to make sure we're within a repo
         os.chdir(os.path.dirname(__file__))
         self.assertIsNotNone(utils.pwd_git_version())
-        with patch('shub.utils.find_executable', return_value=None):
+        with patch('shub.utils.which', return_value=None):
```
wRAR, line 77: Should this be `shub.utils.which` or `shitil.which`?

shub/utils.py
branchvincent: Restored backward compat, ended up moving `keep_log` out of this function anyway since it was simpler. also updated the tests to actually assert what was logged.

is a deprecation warning still necessary? I assumed this function was an internal utility and not user-facing

tests/test_utils.py
branchvincent: This confused me at first too, but because `shub.utils` is already imported before applying the `patch` we have to use the [same object name](https://docs.python.org/3/library/unittest.mock.html#where-to-patch
) that `pwd_git_version` refers to (other tests also do this, see `@patch('shub.utils.time.time')` right below)

tests/test_utils.py
```diff
@@ -274,46 +275,46 @@ def jri_result(follow, tail=None):
     def test_latest_github_release(self, mock_get):
         with self.runner.isolated_filesystem():
             mock_get.return_value.json.return_value = {'key': 'value'}
-            self.assertDictContainsSubset(
-                {'key': 'value'},
-                utils.latest_github_release(cache='./cache.txt'),
+            self.assertLessEqual(
+                {'key': 'value'}.items(),
+                utils.latest_github_release(cache='./cache.txt').items(),
```
Gallaecio, line 280: TIL, nice.

tests/test_utils.py
```diff
@@ -382,20 +383,19 @@ def _call(*args, **kwargs):
...
     def test_echo_short_log_if_deployed(self):
-        log_file = Mock(delete=None)
         last_logs = ["last log line"]
 
         deployed = True
-        for verbose in [True, False]:
-            utils.echo_short_log_if_deployed(
-                deployed, last_logs, log_file, verbose)
-            self.assertEqual(None, log_file.delete)
+        for verbose, expected in zip([True, False], ["", "last log line\n"]):
```
Gallaecio, line 389: :lipstick: For static values, `zip` feels overkill.

```suggestion
        for verbose, expected in ((True, ""), (False, "last log line\n")):
```

Gallaecio (approved):

tests/test_utils.py
branchvincent: good point!

wRAR (approved):
Thanks!

PR #432, opened 2023-10-30
use app.zyte.com now

HassanQamar07:
Resolves https://github.com/scrapinghub/shub/pull/432.

PR #436, opened 2023-12-27
fix tests for latest pipenv

elacuesta:
Taken from #434:

> Failing on master for pipenv>2023.2.4, which was the last version supported for python<3.8

Splitting so we can release #432 sooner.


Gallaecio (approved):

elacuesta:
Thank you @branchvincent!

PR #440, opened 2024-01-18
Drop support for <3.8

### LINKED ISSUE #406: drop support for Python 2.7, 3.5, 3.6
[Python 2.7](https://www.python.org/dev/peps/pep-0373/) reached end of life on January 2020.
[Python 3.7](https://www.python.org/dev/peps/pep-0537/) reached last bugfix release on June 2020
[Python 3.8](https://www.python.org/dev/peps/pep-0569/) had its last bugfix release on May 2021
[Python 3.9](https://www.python.org/dev/peps/pep-0596/) will have its last bugfix release on May 2022
[Python 3.10](https://www.python.org/dev/peps/pep-0619/) is due on October 2021, and will correctly handle code  since 3.7

branchvincent:
Follow up to https://github.com/scrapinghub/shub/pull/438#issuecomment-1881769015 along with CI simplification and an explicit dep on `setuptools` for 3.12

Closes #406


.github/workflows/tests.yml
```diff
@@ -9,130 +9,32 @@ on:
...
+  tests:
+    name: "Test: py${{ matrix.python-version }}, ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}-latest
     strategy:
       fail-fast: false
       matrix:
-        include:
-        - python-version: '3.6'
-          os: ubuntu-20.04
-          tox-env: min
```
Gallaecio, line 21: This env is no longer executed now.

tox.ini
```diff
@@ -10,10 +10,6 @@ deps =
 commands =
     pytest --cov=shub --cov-report=term-missing --cov-report=html --cov-report=xml {posargs:shub tests}
 
-[testenv:min]
-deps =
-    {[testenv]deps}
```
Gallaecio, line 15: Oh, I see this was the same as `py`.

Ideally we should have a `min` env with pinned minimum versions of dependencies, but since it was not there, I’m OK with removing it and adding it back properly later in a separate PR.

Gallaecio (approved):

PR #442, opened 2024-01-22
Address lacking importlib metadata in Python ≤ 3.9

Gallaecio:


Gallaecio:
The CI thing is a bit dirty, but in addition to checking for a regression on this specific issue, it also showed that we had an undeclared dependency on setuptools that would be an issue in Python 3.12.

wRAR (approved):

kmike (approved):

PR #453, opened 2025-03-03
Fix VCS pip line entry in test suite

wkunert:
Since pipenv version 2024.3.0, an additional whitespace character is placed between the package name and the @ character (see https://github.com/pypa/pipenv/pull/6282).


kmike (approved):

Gallaecio:
Thanks!

PR #454, opened 2025-03-05
Add support for current Poetry versions

wkunert:
The previous lockfile parser did not work reliably for Poetry 1.2 and later due to the new lockfile format (e.g., it did not take markers into consideration). This solution should work for all Poetry versions since it uses the `poetry export` command to generate a `requirements.txt` file.

I encountered an issue where shub deploy failed with a requirement error because it attempted to install `pywin32`. This happened because the marker `sys_platform == "win32"` was missing from the generated requirements.txt file.

I’m not sure if it still makes sense to test the conversion from poetry.lock to requirements.txt, as Poetry's test suite already covers this extensively. I adapted the test, but in my opinion, it could be dropped.

The biggest caveat of the new implementation is that users need to have Poetry installed. However, I think this is a safe assumption for this scenario. Or should the old implementation serve as a fallback if the `poetry` command is not found?

Let me know if supporting current Poetry versions is desirable but this solution is not sufficient—I’m happy to refine it or explore alternative approaches.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   tests-ubuntu:
     name: "Test: py${{ matrix.python-version }}, Ubuntu"
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os || 'ubuntu-latest' }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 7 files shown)


wkunert:
I will fix or drop the failing tests as soon as you have decided if and how you want to proceed. 🙂

wRAR (approved):

wkunert:
Thanks for taking care @Gallaecio! 🙏

PR #457, opened 2025-04-18
feat: bumps image size limit to 4GB

alexmarie78:
This should help us deploy a bigger image on zyte.

### DIFF
--- a/shub/image/test.py
+++ b/shub/image/test.py
@@ -1,7 +1,6 @@
 import click
-
 from shub import exceptions as shub_exceptions
-from shub.config import load_shub_config, list_targets_callback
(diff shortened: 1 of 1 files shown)


shub/image/test.py
```diff
@@ -19,38 +18,48 @@
 in its end (if you do not provide -S/--skip-tests parameter explicitly).
 """
 
-IMAGE_SIZE_LIMIT = 3 * 1024 * 1024 * 1024  # 3GB
+IMAGE_SIZE_LIMIT = 4 * 1024 * 1024 * 1024  # 4GB
```
alexmarie78, line 21: Real change here, others are formatter-related

PR #468, opened 2026-04-10
Prioritize env vars over files

AdrianAtZyte:
cc @VMRuiz


wRAR (approved):

kmike:
Hey! Coud you elaborate, why is it a desired behavior?

VMRuiz:
Hello @kmike,

There is a common situation where a single user has access to multiple projects, for example Project A and Project B. Since the same account is authorized to deploy to both, they could accidentally deploy Project B’s code to Project A.

Using project-specific environment credentials per folder would help reduce that risk. Each project folder could activate only the credentials needed for that specific project, making accidental cross-project deployments much less likely.

PR #476, opened 2026-06-17
Support for reading SHUB_APIKEY from .env file

elacuesta:
Since we only care about `SHUB_APIKEY` this PR uses `dotenv_values` instead of `load_dotenv` to avoid overriding other keys.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -130,7 +130,8 @@ file and let ``shub`` load it:
 
     SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
 
-By default ``shub`` reads the ``.env`` file in the current directory. Use the
+By default ``shub`` reads the nearest ``.env`` file, looking in the current
+directory and then walking up through its parent directories. Use the
 ``--dotenv-path`` option to point it at a different file::
 
     shub --dotenv-path /path/to/myenv deploy
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -123,7 +123,10 @@ On Windows::
     SET SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
 
 Instead of exporting the variable yourself, you can store it in a ``.env``
-file and let ``shub`` load it::
+file and let ``shub`` load it:
+
+.. code-block:: bash
+    :caption: :file:`.env`
 
     SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -122,6 +122,20 @@ On Windows::
 
     SET SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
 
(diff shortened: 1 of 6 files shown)


docs/configuration.rst
```diff
@@ -122,6 +122,20 @@ On Windows::
 
     SET SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
 
+Instead of exporting the variable yourself, you can store it in a ``.env``
+file and let ``shub`` load it::
+
+    SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
```
AdrianAtZyte: 💄 

```suggestion
file and let ``shub`` load it:

.. code-block:: bash
    :caption: :file:`.env`

    SHUB_APIKEY=0bbf4f0f691e0d9378ae00ca7bcf7f0c
```

pyproject.toml
```diff
@@ -18,6 +18,7 @@ dependencies = [
 	'importlib-metadata; python_version < "3.10"',
 	"packaging",
 	"pip",
+	"python-dotenv>=1.0.0",
```
AdrianAtZyte, line 21: Maybe the `min` env in tox.ini should freeze this version?

AdrianAtZyte:
~~(the test error looks like a bad expectation to me, the code looks good to me)~~ Wait…

AdrianAtZyte (approved):

shub/tool.py
```diff
@@ -23,10 +25,29 @@
...
+    """Load SHUB_APIKEY from a .env file into the environment.
+
+    Only the SHUB_APIKEY variable is read from the file; any other variables are ignored.
+    A SHUB_APIKEY already present in the environment takes precedence over the value in
+    the file. When ``dotenv_path`` is None, python-dotenv looks for a ``.env`` file in
+    the current directory.
+    """
+    if 'SHUB_APIKEY' in os.environ:
+        return
+    apikey = dotenv_values(dotenv_path).get('SHUB_APIKEY')
```
AdrianAtZyte: Seems like `dotenv_values()` does not use the cwd, and does not even allow changing that (there is a parameter to use it for the `find_dotenv()` function it calls, but it does not expose that parameter. That must be what is causing the test error.

elacuesta: Indeed, there seems to be a difference in the way that the coverage package is implemented in py3.14, which causes `sys.gettrace` to return `None` [here](https://github.com/theskumar/python-dotenv/blob/v1.2.2/src/dotenv/main.py#L354).
A solution is to lookup the `.env` file explicitly by default 1ef6380

tests/test_tool.py
AdrianAtZyte, line 1: Should we test a case where the .env is in a parent folder? I believe we want that to work, but I wonder if by specifying ".env" that could be problematic (although it probably isn’t), so it may be worth testing just in case.

elacuesta: Good catch 😄  30b7cbe

pyproject.toml
elacuesta: Done 73ec481

AdrianAtZyte (approved):

PR #478, opened 2026-06-29
Remove unused freeze and release actions

### LINKED ISSUE #467: Get rid of the old release action
I noticed there is a "release" action that is not working anymore, and we already have a working "publish" action, so we should remove the release one.

AdrianAtZyte:
Fixes #467.


wRAR (approved):

elacuesta:
For the record these jobs were not working because the actions they were using were no longer supported, however they would work again after #477.

AdrianAtZyte:
Do you think they *should* work, i.e. do you see value in maintaining them?

elacuesta:
I don't have a strong opinion really, I think it comes down to how much do we value the standalone binaries. I haven't ever used them myself, but I don't how widely used they are in general.

elacuesta:
The last release with downloads is 2.5 years old and nobody has complained about the lack of standalone binaries: https://github.com/scrapinghub/shub/releases/tag/v2.15.2, we probably don't need to worry about it.

PR #480, opened 2026-07-16
Exclude gitignored files from shub deploy by default

alexldide:
- _**shub deploy**_ used to build the egg straight from the working directory, so gitignored files (secrets, build artifacts, venvs) could leak into a deploy. Now, inside a git repo, the egg is built from a filtered copy that excludes anything git ignores (git ls-files --cached --others --exclude-standard), uncommitted edits and new untracked-but-not-ignored files are still included. No new flag; non-git projects are unaffected.                                                                                                                                               

  
 Jira ticket: https://zyte.atlassian.net/browse/DC-902


AdrianAtZyte:
> Default behavior (shub deploy without the flag/option) is unchanged

This feels like a bug fix to me, so I wonder if we should change the default behavior, and use the old behavior only if Git is not being used or if some new flagged is passed to force the old default behavior.

AdrianAtZyte:
@kmike @wRAR @elacuesta Thoughts?

wRAR:
Some unstructured thoughts:

1. To me the change description ("(build artifacts, local secrets, stray virtualenvs, etc.) never end up in the deploy") 100% sounds like the desired behavior, whatever the actual implementation of that. 
2. I'm not sure about using git status/git checkout for that but that may be the only automatic way to do anything, others would need allow/deny lists. I know that many packaging tools have allow list (hatchling), deny lists (vsce with .vsceignore), use git status (hatch-vcs?), use .gitignore or package the whole folder by deafult (setuptools IIRC). Though just using `.gitignore` may be less disruptive for the default behavior.
3. It may be surprising to users to have this on by default as I think deploying temporarily uncommitted changes must be very popular.

alexldide:
@AdrianAtZyte, @wRAR 
So the default (non-flag) behavior should be "build from the working tree as it is, but exclude anything git already flags as ignored/untracked-ignored" but keep uncommitted edits to tracked files?

AdrianAtZyte:
Yes, I think that makes sense.

In fact, I wonder if that would be what you actually were looking for, or if you really need the Git HEAD to be deployed ignoring local changes. I would expect CI deploys to be the ones needing the HEAD, and those not needing the flag because they are always clean, and your local runs will probably either be meant to include local changes as @wRAR points out, or you could probably "git stash; shub deploy; git stash pop" instead of needing a new flag.

alexldide:
Seems like a dead end for clean repo flag. I'll revert the implementation and build the gitignore-filtered default instead, same working tree, just excluding whatever git already flags as ignored.

AdrianAtZyte (approved):

